PR #17, opened 2019-06-27
HID-1663 - update all list counts when a user is deleted

guillaumev:


api/models/User.js
```diff
@@ -682,10 +682,26 @@ UserSchema.pre('remove', async function (next) {
...
+    if (!this.authOnly && !this.hidden) {
+      if (!this.is_orphan && !this.is_ghost) {
+        updates.countManager = -1;
+        updates.countVerified = -1;
+        updates.countUnverified = -1;
+      } else {
+        updates.countManager = -1;
+        updates.countVerified = -1;
+      }
+    }
```
rupl: So is the idea that we're manually adjusting some values when performing operations that affect the counts? I ask for my own understanding.

guillaumev: Yes correct. In this case, when a user is deleted from HID (or deletes his own account), the list counters of the lists the user is checked into are updated to reflect this.

guillaumev:
@rupl ok to merge ?

rupl:
lgtm 👍

PR #18, opened 2019-06-27
HID-1804 - Added mailhog to docker-compose example file and nodemon

guillaumev:
We will have to see with @teodorescuserban if it's possible to run "npm start" on production and dev, but run "npm run dev" on our local environments to have it run with nodemon...

### DIFF
--- a/docker-compose.example.yml
+++ b/docker-compose.example.yml
@@ -5,9 +5,6 @@ services:
     image: unocha/mongo:4.0.5
     volumes:
       - ./db:/srv/db
-    # environment:
-    #   DNSDOCK_NAME: db
-    #   DNSDOCK_IMAGE: hid
   dev:
     image: unocha/nodejs:8.11.3
     volumes:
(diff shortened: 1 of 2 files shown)


rupl:
I do believe it's possible. For example on DSR we have a [`run_node.dev`](https://github.com/UN-OCHA/reports-site-stack/blob/master/env/local/run_node.dev) file in the stack which is intended for local development, but the infrastructure runs this [`run_node` file in the DSR repo](https://github.com/UN-OCHA/reports-site/blob/dev/run_node)

docker-compose.example.yml
```diff
@@ -5,46 +5,24 @@ services:
...
-      SMTP_HOST: app.mailhog.vm
+      ROOT_URL: http://localhost:3000
+      SMTP_HOST: mailhog
       SMTP_PORT: 1025
-      SMTP_USER: humanitarianid
-      SMTP_PASS: smtp_password
-      APP_URL: http://app.hid.vm
+      SMTP_USER: 
+      SMTP_PASS:
+      APP_URL: https://app.hid.vm
```
teodorescuserban: what is this for?

teodorescuserban:
Indeed, you map your custom startup script as `/etc/services.d/node/run`.

docker-compose.example.yml
guillaumev: Some emails sent out by the API have a link which needs to point to the app URL, which is configured here

guillaumev:
@rupl can you check this and make sure it doesn't break things on your side ?

docker-compose.example.yml
```diff
@@ -5,46 +5,24 @@ services:
...
       NODE_ENV: local
-      ROOT_URL: http://api.hid.vm
-      SMTP_HOST: app.mailhog.vm
+      ROOT_URL: http://localhost:3000
+      SMTP_HOST: mailhog
       SMTP_PORT: 1025
-      SMTP_USER: humanitarianid
-      SMTP_PASS: smtp_password
-      APP_URL: http://app.hid.vm
+      SMTP_USER:
```
emmajane: Does this intentionally remove a user/pass?

teodorescuserban: it makes possible to actually use the SMTP_USER and SMTP_PASS in the `.env` file to define it for the desired environment. Providing a hardcoded default was useless anyway.

PR #21, opened 2019-07-11
HID-1817: update libphonenumber

rupl:
## https://humanitarian.atlassian.net/browse/HID-1817

Blocks https://github.com/UN-OCHA/hid_app2/pull/147

Without this update, the client-side library can validate but then the visitor is shown an error when clicking the <kbd>+</kbd> button.


rupl:
Functional test was successful on my local. Not much else to it I think. 👍

PR #22, opened 2019-07-16
HID-1806 - user profiles show same auth client multiple times

guillaumev:
Changed the code according to the comment in HID-1806, which I agree improves the code, but I'm not sure this will fix the issue...

### DIFF
--- a/api/controllers/AuthController.js
+++ b/api/controllers/AuthController.js
@@ -364,7 +364,10 @@ module.exports = {
       request.auth.credentials = user;
       // Save authorized client if user allowed
       const clientId = request.yar.authorize[request.payload.transaction_id].client;
-      if (!request.payload.cancel && !user.hasAuthorizedClient(clientId)) {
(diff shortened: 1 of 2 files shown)


api/controllers/AuthController.js
```diff
@@ -364,7 +364,10 @@ module.exports = {
       request.auth.credentials = user;
       // Save authorized client if user allowed
       const clientId = request.yar.authorize[request.payload.transaction_id].client;
-      if (!request.payload.cancel && !user.hasAuthorizedClient(clientId)) {
+      if (!request.payload.bsubmit || request.payload.bsubmit === 'Deny') {
+        return reply.redirect('/');
+      }
+      if (!user.hasAuthorizedClient(clientId) && request.payload.bsubmit === 'Allow') {
```
rupl: 💯

rupl:
Yep no problem as far as I know. I just avoided merging new code in case we had to do emergency fixes while you were gone.

PR #26, opened 2019-08-27
HID-1829 - added nonce and respect scopes

guillaumev:


api/services/JwtService.js
```diff
@@ -46,6 +46,17 @@ module.exports = {
...
     };
+    if (scope.indexOf('email') !== -1) {
+      idToken.email = user.email;
+      idToken.email_verified = user.email_verified;
+    }
+    if (scope.indexOf('profile') !== -1) {
+      idToken.name = user.name;
+      idToken.family_name = user.family_name;
+      idToken.given_name = user.given_name;
+      idToken.picture = user.picture;
```
rupl: Picture isn't a required field. Is there some fallback that is guaranteed to exist or should we provide one with a double-pipe operator?

guillaumev: Good catch, I thought it was actually empty by default. Let me know what you think of this one: https://github.com/UN-OCHA/hid_api/pull/27

PR #27, opened 2019-08-27
HID-1829 - make picture empty as default

guillaumev:


rupl:
oh interesting, I had been focusing on the JwtService side of this but globally setting a default seems even better.

PR #30, opened 2019-09-05
HID-1839 - Improved logging on various controllers and refactored som…

guillaumev:
…e deprecated code


api/controllers/ListUserController.js
```diff
@@ -272,57 +328,11 @@ module.exports = {
     // Synchronize google spreadsheets
     promises.push(GSSSyncService.deleteUserFromSpreadsheets(list._id, user.id));
     promises.push(OutlookService.deleteUserFromContactFolders(list._id, user.id));
+    logger.info(
+      '[ListUserController->checkout] Saving list and user for checkout and sending notifications',
+    );
     await Promise.all(promises);
```
rupl: About these Promises executions: I see we're adding all the logging as various conditions are detected, but it's kind of premature. The actual code which might succeed or fail is accumulated into this `Promise.all`, which has no `then`/`catch` to detect whether something worked or not.

I assume if one of them fails, it 500s the whole chain of events, right? Even though some or many might have already succeeded. That seems to be the behavior I've seen when I debug locally and have some minor config problem (like a missing GSSSync key). It reports failures when nothing critical actually failed.

Maybe it would be better to queue these logging events and only execute them once `Promise.all` succeeds:

```js
const pendingLogs = []; // push the log message each time you push to `promises`
pendingLogs.push({type: 'info', messages: ['something was successful', object]});
pendingLogs.push({type: 'info', messages: ['something else was successful', string]});

await Promise.all(promises).then((res) => {
  pendingLogs.forEach(log => {
    logger[log.type](...log.messages);
  });
}).catch((err) => {
  logger.error('[ListUserController->update] some operations failed during batch execution', err);
});
```

(this is technically pseudocode, it might not work. just to get the idea across)

commands/updateListUsers.js
```diff
@@ -0,0 +1,71 @@
...
+  }
+  process.exit();
+}
+
+(async function () {
+  await run();
+}()).catch((e) => {
+  console.log(e);
+  process.exit(1);
+});
```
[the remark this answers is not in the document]
guillaumev: This command is actually a command that was run once, to reinitialize the checkins of all users, and will (hopefully) never run again. I still want to keep it as a command in case it needs to run again for some reason, but basically, I don't really care if we have to revert this change (as it's just moving code from one place to another, ie moving it from the ListUserController to a command file)

api/controllers/ListUserController.js
guillaumev: I agree when you say that logging when adding the promise to the promise array is probably premature, but at the same time I believe that once all logging messages have been added, if an error happens in, say, GSSSyncService, it will be logged.
Also, in the pseudocode you wrote, the only difference I see with what I'm proposing is that you're writing the log messages after everything gets executed, while I'm writing them before things get executed, and I don't really understand why this would be more helpful for debugging...

rupl: By logging after `Promise.all` succeeds, your logs are verified to be reporting events which actually occurred.

If you log beforehand and your `Promise.all` fails (since it can fail when just one single Promise throws an exception), then any number of things that got logged beforehand might not have actually worked.

I suppose if we take async out of the picture, logging beforehand is like doing this with jQuery:

```js
// premature log before event ever happens
console.log('the document is ready');

$document.ready(function () {
  // the console.log statements should be HERE inside the callback,
  // where we know the document is ready
  //
  // using callback is equivalent to .then() in Promises or async/await
  $('.something').thatRequiresDomReady();
});

// it is definitely not ready here either
console.log('the document is definitely ready now');
```

guillaumev: If you log afterwards, a lot of things could also happen and succeed before you actually get any log, and you won't know what happened because nothing got logged...
If I look back at your pseudocode and assume there is an error so we go into the "catch" block, you will see the error message in the logs, but you won't be able to tell which promises actually tried to execute (and eventually succeeded) before you got to the error message.
So basically, I think either way, the fact that we are executing in an async manner can make logs relatively messy, because one execution won't lead to the same log trace as another one...

rupl: To sum up our post-standup chat today (myself, @emmajane, @guillaumev), queuing the logs and playing them back inside `Promise.all().then()` has the advantage of avoiding log output during times when known validation errors occur (typically a `logger.warn` followed by `throw Boom`), which would avoid the confusion of logging prematurely, but never actually executing the planned operation.


Additionally, we can add the output of `pendingLogs` to the `.catch()` block as well, in order to indicate what did (or didn't) happen before exceptions that occur during `Promise.all`.

api/controllers/UserController.js
```diff
@@ -287,26 +310,49 @@ module.exports = {
         delete request.payload.tester;
       }
 
+      // Not showing request payload to avoid showing user passwords in logs.
       const user = await User.create(request.payload);
       if (!user) {
+        // Not showing request payload to avoid showing user passwords in logs.
```
rupl: As an alternative, you can always do some string replacement on the sensitive property. See this line in Snap Service where we want to provide debug info, but also want to avoid leaking sensitive data into logs:

https://github.com/UN-OCHA/tools-snap-service/blob/e68e6ffda5cb0d607f290cba4d7a45b4ff175e16/app/app.js#L251

api/controllers/UserController.js
```diff
@@ -475,20 +537,32 @@ module.exports = {
       request.payload.verificationExpiryEmail = false;
     }
     if (request.payload.old_password && request.payload.new_password) {
-      logger.warn('Updating user password', { request, security: true });
+      logger.warn(
```
rupl: Is this warning or info?

api/controllers/UserController.js
```diff
@@ -475,20 +537,32 @@ module.exports = {
...
+          );
           throw Boom.badRequest('Password is not strong enough');
         }
         request.payload.password = User.hashPassword(request.payload.new_password);
         request.payload.lastPasswordReset = new Date();
         request.payload.passwordResetAlert30days = false;
         request.payload.passwordResetAlert7days = false;
         request.payload.passwordResetAlert = false;
-        logger.warn('Successfully updated user password', { request, security: true });
+        logger.warn(
```
rupl: this also seems like info or even debug

api/controllers/UserController.js
```diff
@@ -569,62 +659,98 @@ module.exports = {
...
     if (!user) {
+      logger.warn(
+        `[UserController->destroy] Could not find user ${request.params.id}`,
+      );
       throw Boom.notFound();
     }
     await EmailService.sendAdminDelete(user, request.auth.credentials);
+    logger.info(
+      `[UserController->destroy] Removing user ${request.params.id}`,
+    );
```
rupl: this one precedes the DB operation ;)

you were just testing to see if I read the whole PR, weren't you?

api/controllers/UserController.js
```diff
@@ -839,7 +1032,10 @@ module.exports = {
     record.passwordResetAlert = false;
     record.lastModified = new Date();
     await record.save();
-    logger.warn('Password updated successfully', { security: true, request });
+    logger.warn(
+      `[UserController->resetPasswordEndpoint] Password updated successfully for user ${record._id.toString()}`,
+      { security: true, request },
+    );
```
rupl: this looks like info to me

api/controllers/OperationController.js
```diff
@@ -1,18 +1,23 @@
...
 
 /**
  * @module OperationController
- * @description Generated Trails.js Controller.
+ * @description CRUD controller for operation pages.
  */
 module.exports = {
 
   async create(request) {
     const operation = await Operation.create(request.payload);
```
rupl: no `info` for creating operation?

api/controllers/WebhooksController.js
```diff
@@ -21,13 +22,19 @@ async function _notifyNewDisaster(disaster) {
       /* eslint no-await-in-loop: "off" */
       const list = await List.findOne({ remote_id: operation.id });
       if (!list) {
+        logger.warn(
+          `[WebhooksController->_notifyNewDisaster] List with remote_id ${operation.id} not found`,
+        );
```
rupl: I think this one should be `logger.error`

api/services/EmailService.js
```diff
@@ -46,6 +47,9 @@ function send(options, tpl, context) {
     message: options,
     locals: context,
   };
+  logger.info(
+    `[EmailService->send] About to send ${tpl} email to ${options.to}`,
+  );
```
rupl: `logger.debug` for this one?

api/services/EmailService.js
```diff
@@ -46,6 +47,9 @@ function send(options, tpl, context) {
...
   return email.send(args);
```
rupl: then maybe the `logger.info` should follow successful transmission of the email. I see it's actually `return`ing the result of the email btu maybe it can be `await`ed and then returned after logging? if not, no biggie.

config/routes.js
```diff
@@ -171,15 +171,6 @@ module.exports = [
...
 
-  {
-    method: 'GET',
-    path: '/api/v2/updatelistusers',
-    handler: ListUserController.updateListUsers,
-    config: {
-      auth: false,
-    },
-  },
-
```
rupl: I know we already discussed this, so it's not a request for a modification per se, but I want to go on record saying it seems very unnecessary to mix removals of API code with other PRs (even if you're personally quite confident it won't have an effect.. we all make mistakes sometimes).

If someone's using this, we just very unceremoniously slipped a breaking change into a commit that doesn't even announce the change. it just says "improved logging" on most commits in this PR, with two mentions of "and removed code"

I'd personally isolate API removals in their own commit, PR, and release.

Removing an API method without typically implies a new major version for the software. I assume this removal is part of the deprecated code mentioned in the commits, but do we have usage stats that guarantee it is unused at the moment? I would think this logging PR will give us that visibility, but until it's live we can't collect that data so easily.

api/controllers/UserController.js
guillaumev: According to OICT, if I remember right, they want all sensitive operations to be logged with a relatively high level of logging, which is why if I remember right I used warn instead of info. I'm happy to use info instead but I'm not sure it's OICT compliant...

api/controllers/UserController.js
guillaumev: See my reply concerning OICT...

api/controllers/UserController.js
guillaumev: See my reply concerning OICT

api/services/EmailService.js
guillaumev: logger.debug would mean that it wouldn't be logged in production, and I think it might be interesting to have some logs of the emails that are sent out no ?

api/controllers/UserController.js
emmajane: OICT checkpoints relating to logging :

- All access control decisions can be logged and all failed decisions are logged.
- All input validation failures are logged.
- Security logging controls provide the ability to log both success and failure events that are identified as security-relevant.
- an indication that this is a security relevant event (if mixed with other logs);
- The security logging system includes a log analysis tool which allows the analyst to search for log events based on combinations of search criteria across all fields in the log record format supported by this system.

I think as long as the logs are marked with "security" the log "level" (info/warn/debug/error) isn't specified by OICT.

api/services/EmailService.js
guillaumev: It returns the promise back to the sendX function which itself returns the promise back to the controller, and the controller normally has a log message saying that the email was sent successfully.

api/services/EmailService.js
rupl: I didn't think of that. carry on!

api/services/EmailService.js
rupl: gotcha, thanks for the explanation

config/routes.js
guillaumev: This is not an API method. 
This is a callback which I used at a much earlier point before switching to nodejs commands via Jenkins, which has now been moved to a command. I agree that it would be better if it were in its own commit, but I'm sure no one is using this (and even if someone is, this is not code that should be used anymore anyways).

api/controllers/UserController.js
guillaumev: Should be fixed in my last commit.

api/controllers/OperationController.js
guillaumev: Should be fixed by my last commit

api/controllers/WebhooksController.js
guillaumev: Should be fixed by my last commit.

api/controllers/UserController.js
guillaumev: Ok, changed to info in my last commit.

api/controllers/UserController.js
guillaumev: Changed to info in my last commit.

api/controllers/UserController.js
guillaumev: Changed to info in my last commit.

config/routes.js
rupl: Ok, I can accept that. Maybe in the future we can also document deprecated things and schedule their removal, or at least mark them as needing to be removed at some point. 

Lacking clear documentation seems to be the main problem here, causing me to be overly cautious.

PR #33, opened 2019-09-26
HID-1810 - Upgrade nodeJS to latest version

guillaumev:


rupl:
I don't think this will have an effect on our local environments unless the `docker-compose.yml` is also updated. Until I changed the image that `dev` uses I couldn't get node 10 running locally.

At a minimum we should update the `docker-compose.example.yml` that is tracked by git. Once I changed the docker-compose file, I got results:

```sh
$ doco exec dev node -v
v10.14.2
```

Once I did that, I wiped `node_modules` and after a few rocky restarts during installation, it seemed like smooth sailing with a fully-installed `node_modules`.

Still, I think @teodorescuserban should also look this over and confirm since he knows which files are needed by local devs and which are used by infra.

guillaumev:
@rupl Agreed to change the version on docker-compose.example.yml, I just did it. Also agree that @teodorescuserban should have a look at this, that's why I added him as a reviewer :-)

PR #34, opened 2019-09-27
HID-1853 - fixed configuration of the client application is invalid e…

guillaumev:
…rror message

### DIFF
--- a/api/controllers/ViewController.js
+++ b/api/controllers/ViewController.js
@@ -97,7 +97,7 @@ module.exports = {
         const client = await Client.findOne({ id: request.query.client_id });
         if (!client
           || (client && client.redirectUri !== request.query.redirect_uri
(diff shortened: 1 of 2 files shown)


api/controllers/ViewController.js
```diff
@@ -95,7 +95,9 @@ module.exports = {
     if (request.query.client_id) {
       try {
         const client = await Client.findOne({ id: request.query.client_id });
-        if (!client || (client && client.redirectUri !== request.query.redirect_uri)) {
+        if (!client
+          || (client && client.redirectUri !== request.query.redirect_uri
+            && client.redirectUrls.indexOf(request.query.redirect_uri) === -1)) {
```
rupl: `client.redirectUrls.includes()` would read better to me, but this code is of course valid.

guillaumev: Agree. I changed it.

PR #46, opened 2019-10-21
HID-1822: CD for Auth login page

rupl:
## HID-1822

The AngularJS repo holds all the assets for the branding of the HID website, but one page is served up by the API itself: the Auth login.

This PR updates the HTML/CSS of the Auth page so it is identical as possible to the HID frontend itself. In https://github.com/UN-OCHA/hid_app2/pull/182 we fixed the `sass:api` grunt task so that this process can be smoother/quicker for any future updates.


guillaumev:
@rupl I just pulled this and I have 2 issues (one of them might be a non-issue though):

* Nothing happens when I click on "OCHA Services"
* The footer is not sticky at the bottom (this one might be a non-issue)

rupl:
Ah yes, sorry I didn't mention it here (only in ticket):

- the OCHA Services, per some discussion which I've lost documentation of, is NOT clickable in this context (to avoid distracting people from logging in). The one tiny indication is that the chevron is missing.
- the footer is no longer sticky on Auth or the App itself. We did that intentionally at some point, due to how the angular app was constructed and not wanting to refactor too much.

PR #78, opened 2020-01-28
Allow named arguments

attiks:
Allow named arguments so you can do
```docker-compose exec dev node ./commands/createDummyClient.js --url=https://iasc8.local.docksal/ --redirectUrl=https://iasc8.local.docksal/user/login/hid/callback```


rupl:
ah, actuallly now I see what happened. when I took the exact example you provided, it created an entry with id `client` that was in a different position in the list. I think that's fine since this is local-development only.

PR #81, opened 2020-03-17
HID-1967: enforce policy that new password cannot match old password

rupl:
# HID-1967

While working on HID-2002 I noticed it is possible to craft an API request that "changes" your new password to the current password. All copy on our site/api-docs implies this is not possible. Small conditional is all we needed.

### DIFF
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -889,7 +889,7 @@ module.exports = {
     if (!HelperService.isAuthorizedUrl(appValidationUrl)) {
       logger.warn(
         `[UserController->validateEmail] app_validation_url ${appValidationUrl} is not in authorizedDomains allowlist`,
-        { security: true, fail: true, request },
+        { request, security: true, fail: true },
       );
       throw Boom.badRequest('Invalid app_validation_url');
     }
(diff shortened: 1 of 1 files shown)


rupl:
Ok this got more complicated as I looked into it. Firstly, the AngularJS app is using a wholesale `PUT` to update the password (my first commit on the branch), but there's also a specific password-changing method. To make sure all security-related logs are meeting OICT requirements, I had to reword some stuff to include:

- User ID of any security-related operation involving passwords
- The HID-defined `security` boolean in each log
- General wording changes to make all messages uniform

rupl:
Just so I'm clear, can you confirm that the testing workflow would be as follows:

- Request to reset password via email link
- Use link to reset password
- Once logged in (with the fresh password), now attempt to change password
- pre-patch: password change should NOT be possible
- post-patch: password change should be possible

orakili:
1. Request to reset password via email link
2. Use link to reset password
3. Enter **existing** password as new password in reset password form
    -> pre-patch -> "success" message
    -> post-patch -> "error" message: `The new password can not be the same as the old one`

The patch is just fixing the comparison between new password from reset form and password in DB.

rupl:
Nice catch on some of the logs missing the `security` boolean. I'll add that to every log and enforce some consistency across the objects.

Your patch works as advertised! Added to the branch. I also noted that the user feedback does NOT say that the passwords matched, only that there was an error. By leaving the error message generic, it avoids revealing knowledge of the current password to the visitor 👍

orakili:
@rupl Re. your last point. While the app shows a generic error, the response payload has the message "The new password can not be the same as the old one". I'm a bit worried that could be exploited though one would need to have a valid password reset link for each attempt which limit the risks.

rupl:
Good catch!

I thought I looked through the actual API response but had only checked the headers (that's sometimes where the API leaves important feedback). I see it in the payload. We can open follow-ups to ensure that the API does not reveal the password's value to the user in any way.

rupl:
Actually I just popped it onto the branch since the PR was introducing the security hole. The response is now a generic 'Could not reset password' in the payload when using `resetPasswordEndpoint`. When a user is logged in and trying to change the password, the payload still contains the more specific message.

PR #86, opened 2020-04-10
HID-2016: v3 password strength

rupl:
# HID-2016

Introduce new v3 method for changing user passwords, which has increased complexity requirements.

### DIFF
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -1007,7 +1007,18 @@ module.exports = {
       throw Boom.badRequest('Wrong arguments');
     }
 
-    if (!User.isStrongPassword(request.payload.password)) {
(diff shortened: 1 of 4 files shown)


rupl:
Thanks for spotting those. Yes, the v3 only uses the `.../password` specific endpoint. My intention is to never use a generic `PUT` to update the password. I think it's very out of place to have the extra code inside the `update()` method and it should be removed. But doing so would have required updates to AngularJS so for the time being I left it there.

The `resetPasswordEndpoint` endpoint should also have a matching `/api/v3` so I'll put that in this branch.

The other two are a bit more complicated since they involve breaking changes. I'll file a detailed bug explaining the scope of those changes.

PR #85, opened 2020-04-09
HID-2021 / HID-2022: emails to notify auth/profile users about decommissioning

rupl:
# HID-2021 + HID-2022

Two new emails to notify users about HID v2 decommissioning. There are separate emails for profile and auth users.

Bonus perk of @orakili reviewing is he can spot all the french typos I probably made 😆 

Email subject lines for both EN/FR pending

PR #88, opened 2020-04-15
HID-2021: more french corrections

rupl:
See https://github.com/UN-OCHA/hid_api/pull/85#discussion_r408516235


rupl:
@orakili feel free to merge this if it's correct. thanks!

PR #91, opened 2020-04-20
HID-2035: allow picture uploads to AWS/local environments

rupl:
# HID-2035

We couldn't upload photos on AWS domains. This band-aids a fix.

Updated the list of TODOs in HID-1894 to include this list.


rupl:
@orakili I totally agree, and all the things you describe are captured in HID-1894.

We're just unblocking ops as rapidly as possible at the moment so they can hit the May 11 date. I'll link to your comment in the other ticket.

PR #94, opened 2020-04-24
HID-2038: remove emails using dedicated method

rupl:
# HID-2038

I uncovered this while doing something else but it's worth tending to. This allows us to fix up yet another instance of our clients using generic `PUT`s to do destructive operations.

The substantive change is the addition of parameter validation. The method could not function before, because the parameter was named but not validated.

### DIFF
--- a/config/routes.js
+++ b/config/routes.js
@@ -543,6 +543,7 @@ module.exports = [
       validate: {
         params: Joi.object({
           id: Joi.string().regex(objectIdRegex),
+          email: Joi.string().email({ minDomainSegments: 2 }),
         }),
       },
     },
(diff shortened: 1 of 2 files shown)


config/routes.js
```diff
@@ -543,6 +543,24 @@ module.exports = [
       validate: {
         params: Joi.object({
           id: Joi.string().regex(objectIdRegex),
+          email: Joi.string().email({ minDomainSegments: 2 }),
```
rupl, line 546: @orakili this is the one absolutely necessary line of the PR. The rest was just some optional cleanup stuff I threw in. Avoiding hoisting and all that..

PR #96, opened 2020-04-29
HID-2035: add another domain for AWS migration

rupl:
# HID-2035

New environment, new hardcoded domain.


rupl:
@teodorescuserban FYI. Maybe we could try hotfixing the container and see if this works?

teodorescuserban:
erm, yes it does, thats how i managed to have the reset password form to work.

PR #105, opened 2020-06-08
HID-2060: viewable swagger docs + auth + JWT tweaks

rupl:
# HID-2060

PR adds HTML/CSS/JS assets to actually view the docs. Previously we were producing the underlying YML but not processing it into a usable website. The docs can now be viewed at http://localhost:3000/docs/v3/

The other foundational aspect is that I figured out the Auth header. It's applied as a global default, so all new methods have it automatically. The one existing exception is POST `/jsonwebtoken` which is how a JWT is generated. The requirement is unset on that method, and I documented how to do it for the other methods to be documented that cannot have an Auth header to function properly.

Pictured is the docs page once the Auth token is entered. Notice the POST method is missing a lock icon. Almost all other methods will receive this lock icon by default.

swagger-v3-docs-auth

Additionally, v3 routes for JWTs were created. They point to the same callbacks because no additional work was required for v3.


rupl:
it's ok, I will be using it on dev frequently enough to test it.

PR #106, opened 2020-06-17
HID-2060: v3 user methods

rupl:
# HID-2060

Big docs PR, plus some other minor fixes that either needed to happen to avoid breaking things, or were small enough that I just didn't feel like waiting:

- HID-2061
- HID-2063

Other tickets filed during the course of this work will be ready for review separately.

I did rearrange the order of some endpoints because I personally felt like it was easier to scan when I read the docs page. All sub-paths of `/api/v3/user/{id}` are grouped together. Because of how `swagger-inline` works it meant I had to move the functions themselves inside `UserController.js` — I looked for a way to apply weights to the definitions but they don't seem to exist. Anyway, it doesn't affect the DX at all IMO, but scanning docs makes a little more sense.

Comparing to https://dev.api-humanitarian-id.ahconu.org/docs/#/ will reveal that many things are missing. To the best of my knowledge, anything omitted in the v3 docs is because it won't be available anyway, so I cut my work by ignoring things we're planning on deprecating.

### DIFF
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -871,7 +871,7 @@ module.exports = {
       logger.warn(
         `[UserController->destroy] User ${request.auth.credentials._id.toString()} is not allowed to delete user ${request.params.id}`,
       );
-      return reply(Boom.forbidden('You are not allowed to delete this account'));
+      throw Boom.forbidden('You are not allowed to delete this account');
     }
 
     const user = await User.findOne({ _id: request.params.id });
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -1278,6 +1278,49 @@ module.exports = {
     throw Boom.badRequest('No file found');
   },
 
+  /*
+   * @api [post] /user/{id}/emails
+   * tags:
+   *   - user
+   * summary: Add a new email to the user's profile.
+   * parameters:
+   *   - name: id
+   *     description: A 24-character alphanumeric User ID
+   *     in: path
+   *     required: true
+   *     default: ''
+   * requestBody:
+   *   description: Email address and validation URL.
+   *   required: true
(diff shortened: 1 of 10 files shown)


api/controllers/UserController.js
```diff
@@ -367,6 +407,75 @@ module.exports = {
...
+   *     in: query
+   *     type: integer
+   *     default: 0
+   *   - name: limit
+   *     description: Maximum number of users to return
+   *     in: query
+   *     type: integer
+   *     default: 50
+   *   - name: fields
+   *     description: Fields to be returned (leave empty to return all)
```
orakili, line 432: Maybe indicate the string format, ex: comma separated list of fields.

rupl:
Regarding TOTP being selectively applied: without any docs or comments, my guess is that it has been applied to "sensitive" fields that directly affect login: generating JWT, password changes, and changing primary email, etc.

Other data (e.g. organization) don't have the potential to disrupt the user's ability to login. I'd say it's fairly normal if we think about GitHub for example: my TOTP or security key is needed at login, and when changing security settings, but changing my profile or other non-essential settings happens without 2FA being required.

emmajane:
90% sure @rupl is correct in terms of field sensitivity. This could be cross-referenced against OICT guidelines to confirm.

api/controllers/UserController.js
rupl: You know, I realized while looking at this that the function shouldn't even exist in v3. We won't be providing the ability to look at another user. Even admins cannot do it anymore.

I'll remove the whole doc block and route.

rupl:
One clue in the codebase is the `security: true` parameter found in logging calls. These are also mandated by OICT and they seem to match up with the same list of operations I mentioned before

PR #114, opened 2020-06-25
HID-2081: DELETE /client/{id} should 404 when client doesn't exist

rupl:
# HID-2081

The operation to delete an OAuth client returned 204 no matter what happened inside the function. That meant successive calls to `DELETE /api/v3/client/{id}` would yield 204 every time. Now it does a lookup and 404s when nothing was found.

I updated the shared v2/v3 function but since this is an admin-only feature of HID, it's not bothering anyone to change behavior.

### DIFF
--- a/api/controllers/ClientController.js
+++ b/api/controllers/ClientController.js
@@ -210,10 +210,19 @@ module.exports = {
    *#     description: Requested client not found.
    */
   async destroy(request, reply) {
-    await Client.remove({ _id: request.params.id });
-    logger.info(
-      `[ClientController->destroy] Removed client ${request.params.id}`,
-    );
-    return reply.response().code(204);
+    const client = await Client.findOne({ _id: request.params.id });
+
+    // If a client was found, delete it and log the event.
+    if (client) {
+      await client.remove();
+
+      logger.info(
+        `[ClientController->destroy] Removed client ${request.params.id}`,
+      );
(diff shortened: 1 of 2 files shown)


docs/v3/hid.yml
```diff
@@ -340,7 +342,7 @@ paths:
         - name: X-HID-TOTP
           in: header
           description: The TOTP token. Required.
-          required: false
+          required: true
```
rupl, line 345: these booleans are all from the `ClientController` docs PR #109

api/controllers/ClientController.js
```diff
@@ -206,14 +206,23 @@ module.exports = {
...
+      await client.remove();
+
+      logger.info(
+        `[ClientController->destroy] Removed client ${request.params.id}`,
+      );
+      return reply.response().code(204);
+    }
+
+    // If no client was found, return 404.
+    return reply.response().code(404);
```
orakili, line 226: Line 132: https://github.com/UN-OCHA/hid_api/blob/de716e4a04d819776110ba5b2102c144b68d0acf/api/controllers/ClientController.js#L132, `throw Boom.notFound()` is used to return a 404. Just wondering: why the difference?

rupl: because I copied the line of code for the 204 and changed it 😁  I'll fix this up.

PR #120, opened 2020-07-03
HID-2093: log payload during failed user creation

rupl:
# HID-2093

There was a comment in the API to not log an entire payload because the password was inside it. Far better to delete that property and send along any other info we have, since the event is logging a validation failure and people might want to debug or otherwise investigate.

While testing I found it in the security policy too.

I also found that I'd forgotten to document the `password` and `confirm_password` params in our new docs. Plus, I noticed that our v3 is also sending the wrong field (`password_confirm` instead of `confirm_password`) so that's a bug that might have otherwise sat for a while before being found.

### DIFF
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -284,17 +284,32 @@ module.exports = {
    */
   async create(request) {
     if (!request.payload.app_verify_url) {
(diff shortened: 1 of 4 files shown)


rupl:
Nice catch! I'll fix it up this morning.

PR #126, opened 2020-07-15
HID-2088: remove links from all email footers

rupl:
# HID-2088

Small follow-up to remove social links from all email templates. They share common header/footer files so this will apply to any email we send.


rupl:
My local mailhog:

Screenshot 2020-07-15 14 31 50

PR #128, opened 2020-07-17
OPS-6812: The Great Log Reformat Pt. 1

rupl:
# OPS-6812

Reshuffling all our logs into a new format as we do the team-wide log refactor. A list of improvements:

- Many logs are changed only in terms of coding style. I defined some conventions in our `CONTRIBUTING.md`
- The log formatter is no longer a copy-pasted function. Every environment uses the same shared formatter now.
- Passing the request actually does something useful now. We pass URL, querystring, headers, JSON payload, auth status... but it's not too verbose — we strip out passwords, and truncate both OAuth client secrets + JWT sigs (preventing someone from using API keys after seeing in ELK).
- The formatter tries to auto-populate the User ID several times based on what data is passed. If you explicitly pass `user.id` then it will not overwrite.
- Removed several fields that seem to be populated by the stack itself from the (previously known as `json` but as of today known as `hid`) object: `os`, `environment`, `hostname`, `@timestamp`
- There's a new `user` object which can be filled with a few standard props: `id`, `email`, `admin` — any other ones are allowed too if you decide to pass it.
- There's a new `oauth` object which can have some common props: `client_id`, `redirect_uri`
- `stack_trace` for stack traces (errors only)


emmajane:
I am inappropriately excited about these changes. 🎉

rupl:
@orakili bumpity bump

api/controllers/UserController.js
```diff
@@ -1593,23 +1884,32 @@ module.exports = {
         }
       }
       const promises = [];
-      const pendingLogs = [];
-      promises.push(record.save());
-      pendingLogs.push({
-        type: 'info',
-        message: `[UserController->validateEmail] Saved user ${record.id} successfully`,
-      });
+      promises.push(record.save().then(() => {
```
orakili, line 1887: Not related to this PR but I have a concern here. The calls to `save`, `sendPostRegister` are not synchronized and `sendPostRegister` is not dependent of the success of `save`. The post_register email may be sent even if saving the user record fails.

api/controllers/UserController.js
```diff
@@ -2014,15 +2384,30 @@ module.exports = {
       user,
     };
     await Promise.all([
-      user.save(),
-      NotificationService.send(notification),
+      user.save().then(() => {
```
orakili, line 2387: Same as above, the `connection_request` email may be sent even if the `user.save()` fails.

config/logs.js
```diff
@@ -0,0 +1,114 @@
...
+    // Log IP if we have it.
+    if (logObject.request && logObject.request.info && logObject.request.info.remoteAddress) {
+      metadata.ip = localLogObject.request.info.remoteAddress;
+    }
+    if (logObject.request && logObject.request.headers && logObject.request.headers['x-forwarded-for']) {
+      metadata.ip = localLogObject.request.headers['x-forwarded-for'];
+    }
+
+    // Include custom user object from log
+    metadata.user = localLogObject.user || {};
```
orakili, line 22: Not sure if the `user` object can contain passwords or other sensitive things that should not be logged? Wouldn't it be safer to sanitize it like done for the `request` below?

config/logs.js
```diff
@@ -0,0 +1,114 @@
...
+
+    // Include custom oauth object from log
+    metadata.oauth = localLogObject.oauth || {};
```
orakili, line 25: Same question: is it possible that sensitive information may be contained in the `oauth` object?

config/logs.js
```diff
@@ -0,0 +1,114 @@
...
+      }
+      if (metadata.request.payload && metadata.request.payload.old_password) {
+        delete metadata.request.payload.old_password;
+      }
+      if (metadata.request.payload && metadata.request.payload.new_password) {
+        delete metadata.request.payload.new_password;
+      }
+
+      // Sanitize OAuth client secrets
+      if (metadata.request.query && typeof metadata.request.query.client_secret !== 'undefined') {
```
orakili, line 66: I would propose to replace `typeof metadata.request.query.client_secret !== 'undefined'` and the equivalent line 77 and 94 by `typeof metadata.request.query.client_secret === 'string'`.

api/controllers/UserController.js
```diff
@@ -398,36 +403,54 @@ module.exports = {
...
         throw Boom.badRequest();
       }
       logger.info(
-        `[UserController->create] User ${user._id.toString()} created successfully`,
+        `[UserController->create] User created successfully with ID ${user._id.toString()}`,
+        {
+          request,
+          security: true,
+          user: {
+            id: user._id.toString(),
```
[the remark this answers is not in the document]
rupl: For the most part I just duplicated whatever variable was being used before. That's why it's different. I wanted to avoid refactoring stuff as much as possible while adding logs.

To your point: `user._id` property should really not be used whatsoever by us as humans. It's a [machine key of type `ObjectId`](https://docs.mongodb.com/manual/reference/method/ObjectId/) inherent to Mongo that can produce instability were we to shard/replicate our database (identical records would have different `ObjectId` values between DB replicas).

I know this and Serban knows it, but we haven't ever ticketed this bug which is built into the foundation of HID API. I will ticket it and come back here with a link. So the real bug is that we should rarely be referencing `_id` (except in a future situation where we've replicated the DB and want to log failures specific to one replica), and thus we'd never need to `toString()` a value which is already a plain string.

Ping @teodorescuserban to make sure I got the details right before ticketing. Please just gimme 👍 / 👎

api/controllers/UserController.js
```diff
@@ -436,11 +459,25 @@ module.exports = {
...
       throw Boom.badRequest('This email address is already registered. If you can not remember your password, please reset it');
     } else {
       logger.warn(
-        `[UserController->create] The user already exists. id: ${record._id.toString()}`,
+        `[UserController->create] The user already exists with ID ${record._id.toString()}`,
+        {
+          request,
+          fail: true,
+          user: {
+            id: record._id.toString(),
```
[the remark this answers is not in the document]
rupl: Same here with `anything._id.toString()` — it implies a much bigger scope of bug than this PR can handle.

api/controllers/UserController.js
```diff
@@ -436,11 +459,25 @@ module.exports = {
...
+          },
+        }
       );
       throw Boom.badRequest(`This user already exists. user_id=${record._id.toString()}`);
```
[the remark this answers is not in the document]
rupl: It's ok. I thought the same thing while inspecting the function and confirmed it.

This error is only returned when the user is authenticated and known to be an admin. The logic above this block will error out with something more generic if the user doesn't have proper perms.

api/controllers/UserController.js
rupl: haha dope. Great catch and i'll ticket the issue.

We already have the mechanism in place (node.js `async`/`await`) so we'll make these wait on each other and get rid of the whole promise queue. There's backstory here that I don't want to retell so I'll just say I'm happy to make this change and use `async`/`await` as intended.

api/controllers/UserController.js
rupl: geez. well hey the nice thing is the code will be easier to read after using plain async/await

config/logs.js
rupl: ooh good point. it often doesn't, but it depends on how the object was loaded. Definitely worth taking a second look and I'll probably end up doing as you recommend and adding some mandatory auto-sanitizing to the log formatter.

rupl: Actually after looking at the next comment about `oauth` I know this is safe. We never log the entire user object wholesale, so the data inside `user` will normally be the following: id and/or email.

config/logs.js
rupl: No, because this is populated based on what we chose to log. We don't log the client secret, which is the only sensitive piece of data regarding an OAuth client.

api/controllers/ClientController.js
```diff
@@ -37,19 +37,27 @@ module.exports = {
...
     const payload = request.payload || {};
-    const client = await Client.create(payload);
-    if (!client) {
+    const client = await Client.create(payload).catch((err) => {
       logger.warn(
         '[ClientController->create] Could not create client due to bad request',
-        request,
+        {
+          request,
+          stack_trace: err.message,
```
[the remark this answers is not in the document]
rupl: Sharp eye, but yes this is ok. I'm happy to leave a comment in the code why, but in this case I checked the error object from mongo and decided the message was more useful than the full stack trace, since the error is a Mongo validation problem. The stack had lots of extra junk but the `err.message` was actionable in terms of providing debugging context, e.g. "client_id field is not unique. cannot create record"

api/controllers/ClientController.js
```diff
@@ -37,19 +37,27 @@ module.exports = {
...
       );
-      throw Boom.badRequest();
+      return client;
     }
 
-    logger.info(
-      '[ClientController->create] Created a new client',
-    );
-    return client;
+    throw Boom.badRequest();
```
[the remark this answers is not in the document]
rupl: Point taken, but this is an admin-only action, so 100% of affected users are in HID flowdock channel. We'll cross that bridge when we come to it.

In my testing (I do actually create fake clients on my local pretty regularly) it either succeeds in creating a new record using your payload, or it throws an error with the problem.

api/controllers/UserController.js
```diff
@@ -792,50 +872,84 @@ module.exports = {
...
+                countVerified: 1,
+                countUnverified: 1,
+              },
+            }).then(() => {
+              logger.info(
+                `[UserController->update] User ${user._id.toString()} is becoming visible. Updated list counts`,
+                {
+                  request,
+                  user: {
+                    id: user._id.toString(),
```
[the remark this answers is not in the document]
rupl: great catch. thanks!

api/controllers/ClientController.js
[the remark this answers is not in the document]
rupl: Here's an example:

```
E11000 duplicate key error collection: local.client index: id_1 dup key: { : "fake-client5" }
```

api/controllers/UserController.js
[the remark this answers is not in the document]
rupl: HID-2101

api/controllers/UserController.js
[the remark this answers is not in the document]
rupl: HID-2101

api/controllers/UserController.js
rupl: HID-2102

api/controllers/UserController.js
rupl: HID-2102

config/logs.js
rupl: nice. Yeah I guess there's no other primitive that can be received so it's more explicit to check before acting on the data as a string.

config/logs.js
rupl: I suppose we could explicitly forbid it in the formatter, but this object is created by the developer in each logging instance. It's not an object pulled from the DB.

api/controllers/UserController.js
```diff
@@ -1352,38 +1598,52 @@ module.exports = {
...
-    pendingLogs.push({
-      type: 'info',
-      message: `[UserController->addEmail] Successfully sent validation email to ${email}`,
-    });
     for (let i = 0; i < record.emails.length; i += 1) {
       promises.push(
-        EmailService.sendEmailAlert(record, record.emails[i].email, request.payload.email),
+        EmailService.sendEmailAlert(record, record.emails[i].email, request.payload.email).then(() => {
+          logger.info(
+            `[UserController->addEmail] Successfully sent email alert to ${record.emails[i].email}`,
```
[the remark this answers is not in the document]
orakili: Ok, never mind. The loop is using `let i` not `var i` so the problem I was thinking about doesn't occur.

rupl:
rick-away-we-go

PR #132, opened 2020-07-27
[OPS-6849] remove .git from image to decrease the size.

teodorescuserban:


teodorescuserban:
since `COPY` is a different step than `RUN`, just removing the `.git` folder wont actually work. added `.dockerignore` instead :)

rupl:
that's what I was going to recommend 😁

teodorescuserban:
old habits die hard. `.dockerignore` was introduced rather late and most of my use cases downloaded archives from the internet to install it and then that archive needed to go away. So the `rm` was in the same `RUN` step with the downloads.

PR #133, opened 2020-07-27
OPS-6812: TIL the difference between clone and cloneDeep

rupl:
# OPS-6812

Bonus: the comment above the affected line of code is now accurate :)


rupl:
@orakili just FYI because it was a real head scratcher to figure this out

PR #134, opened 2020-07-30
HID-2104: create separate admin-only JWT auth endpoint

rupl:
# HID-2104

We want a convenience method to authorize admins only. Since all endpoints are already covered by various security policies this is truly a convenience method to allow our HID Admin tool to block non-admin logins with a clear 403 response.

### DIFF
--- a/api/controllers/AuthController.js
+++ b/api/controllers/AuthController.js
@@ -248,6 +248,105 @@ module.exports = {
     );
     return { user: result, token };
   },
+  /*
+   * @api [post] /admintoken
+   * tags:
+   *   - auth
+   * summary: Admin-only route to generate a JSON web token (JWT)
+   * parameters:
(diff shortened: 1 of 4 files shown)


api/controllers/AuthController.js
```diff
@@ -248,6 +248,106 @@ module.exports = {
...
+   * security: []
+   */
+  async authenticateAdmin(request) {
+    const result = await loginHelper(request);
+
+    // Before proceeding, check if user has admin perms. This is the main
+    // difference between @authenticate and @authenticateAdmin
+    if (!result.is_admin) {
+      throw Boom.forbidden();
+    }
```
rupl, line 296: @orakili this is the only difference between `authenticate` and `authenticateAdmin` — the rest is a direct copy

PR #135, opened 2020-07-31
OPS-6862: ensure all error logs follow same conventions as info/warning

rupl:
# OPS-6862

- Making sure that all instances of `hid.error` are one type of value (string in this case)
- Moved any error objects to a new `err_object` field.
- Added some `security`/`fail` flags when appropriate


rupl:
Just FYI @cafuego

PR #141, opened 2020-09-23
HID-2110: fix fatal errors during OAuth logins

rupl:
# HID-2110

The logging improvements from #128 were causing problems when sending requests with a Basic Auth header. Some cleanup and a few more filters to properly catch the Basic Auth header and also sanitize when found.

I am fairly certain the problem was introduced in 4ee6824 but since I ended up moving code between files I gave up when trying to bisect.

## Testing

Setup the [Extra-secure authorization process](https://github.com/UN-OCHA/hid_api/wiki/Integrating-with-HID-via-OAuth#extra-secure-authentication-process) to test locally. If the initial request does not 500, then the test is successful.

We are also working on patching some dev sites to have a permanent test mechanism on dev. See HID-2110 and OPS-6839 for details.

### DIFF
--- a/config/logs.js
+++ b/config/logs.js
@@ -66,6 +66,8 @@ module.exports = {
       }
 
       // Sanitize OAuth client secrets
+      //
+      // This will display "00000...00000" in ELK.
       if (typeof metadata.request.query.client_secret === 'string') {
         // display first/last five characters but scrub the rest
         const sanitizedSecret = `${metadata.request.query.client_secret.slice(0, 5)}...${metadata.request.query.client_secret.slice(-5)}`;
@@ -95,6 +97,9 @@ module.exports = {
         let sanitizedJWT = metadata.request.headers.authorization.split('.');
         const buffer = Buffer.from(sanitizedJWT[1], 'base64');
         const asciiJWT = buffer.toString('ascii');
+
+        // Now pop the signature off to neuter the usefulness of the logged JWT
+        // and prevent ELK from storing actionable credentials.
         sanitizedJWT.pop();
         sanitizedJWT = sanitizedJWT.join('.');
         metadata.request.headers.authorization = sanitizedJWT;
--- a/config/logs.js
+++ b/config/logs.js
@@ -67,8 +67,8 @@ module.exports = {
 
       // Sanitize OAuth client secrets
       if (typeof metadata.request.query.client_secret === 'string') {
-        // display first/last three characters but scrub the rest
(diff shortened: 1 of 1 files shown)


config/logs.js
```diff
@@ -66,21 +66,40 @@ module.exports = {
...
+
       // Sanitize JWTs, which can allow anyone to masquerade as this user.
       // We do this by removing the signature at the end, which only our
       // server can generate. That way, we can decode and inspect the payload
       // for debugging purposes, without risking the use of the JWT by devs
       // who have access to ELK.
-      if (typeof metadata.request.headers.authorization === 'string') {
+      if (
+        typeof metadata.request.headers.authorization === 'string'
+        && metadata.request.headers.authorization.indexOf('Bearer') !== -1
```
rupl, line 95: this line was the fix: checking for `Bearer` inside the contents of the auth Header

config/logs.js
```diff
@@ -66,21 +66,40 @@ module.exports = {
...
+      //
+      // This will display "Basic 00000...00000" in ELK.
+      if (
+        typeof metadata.request.headers.authorization === 'string'
+        && metadata.request.headers.authorization.indexOf('Basic') !== -1
+      ) {
+        const sanitizedSecret = `${metadata.request.headers.authorization.slice(0, 11)}...${metadata.request.headers.authorization.slice(-5)}`;
+        metadata.request.headers.authorization = sanitizedSecret;
+      }
+
```
rupl, line 87: This block was added to sanitize the `Basic` auth header when present. Otherwise we'd store secrets in Kibana.

config/logs.js
```diff
@@ -66,29 +66,52 @@ module.exports = {
...
-        const asciiJWT = buffer.toString('ascii');
-        sanitizedJWT.pop();
-        sanitizedJWT = sanitizedJWT.join('.');
-        metadata.request.headers.authorization = sanitizedJWT;
-
-        // Auto-populate requesting user ID from JWT unless user.id was explicitly
-        // passed into the log.
-        if (typeof metadata.user.id === 'undefined') {
-          metadata.user.id = JSON.parse(asciiJWT).id;
+        if (sanitizedJWT.length === 3) {
```
rupl, line 98: finally, this conditional was necessary to prevent HID failures during Drupal-based ["Extra secure auth process" step 2](https://github.com/UN-OCHA/hid_api/wiki/Integrating-with-HID-via-OAuth#step-2--request-access-and-id-tokens), where a non-JWT is being sent as a short-lived API key.

orakili (approved):
Looks good to me.

PR #152, opened 2020-10-15
HID-2099 profile

left23:
I copied from existing code to get the route for Profile and Settings, and the method for the handler, which includes the user data.
When I visit the /profile path, I get the correct template and my data when I'm logged in.

Profile and Settings pages have placeholder text. Setting page has tabs.

hid-profile

hid-settings

I've used the cd-dropdown.js for the dropdowns (OCHA services, user menu) so adjusted the markup to accommodate. I also updated selectors and hotlinked the demo site styles as a temp workaround before I sort out the styles properly. The CD version is old and most elements use cd- as prefix for their selectors.

The user menu includes 3 items - Profile, Settings, Log out.

I've split CSS into separate files:

1. cd.css comes from the common_design repo compiled CSS
2. bs.css I think is bootstrap (or another framework) rules and I've commented alot of it out for now, to see what breaks - I suspect it's used for alerts, forms, and maybe some layout
3. normalise.css from Drupal core
4. main.css has colour scheme with CD override rules, and other app-specific rules (maybe some rules that are bootstrap-related)


templates/header.ejs
```diff
@@ -7,6 +7,7 @@
     <title>Humanitarian ID</title>
     <meta name="description" content="Authentication page for Humanitarian ID">
     <meta name="viewport" content="width=device-width, initial-scale=1">
+      <link rel=stylesheet href="https://demo.commondesign-unocha-org.ahconu.org/themes/contrib/common_design/css/styles.css">
```
left23, line 10: @rupl FYI needs auth before it will load.

left23:
@rupl I've ignored the styling thus far, except to hotlink the CSS from the demo site to get the dropdowns working.

templates/header.ejs
left23: No longer relevant, as I merged #154 to this branch

left23:
> @rupl I've ignored the styling thus far, except to hotlink the CSS from the demo site to get the dropdowns working.

I've merged PR #154 so this should be easier to review without the hotlinked CSS file.

rupl (changes requested):
This is a really great start, nice job! Since we are cleared to incrementally merge this into dev, there are a couple aspects I'd like to tweak before we merge this chunk of work:

- The main content column is quite skinny and it forces the Settings tabs onto two rows. You think we can expand the column to take up maybe two more of the Bootstrap grid columns?
- I see you re-used the focus styles on the Settings tabs, but they result in the buttons shifting as I click between them. Could we try using `outline` or at least a transparent border on the non-active links to keep them more steady?
- Maybe this is just me, but I would reverse the order of the fields on Profile editing to be Given, then Family name.

rupl:
Oops just one more thing. I think we can drop "General" and have Password be the default Settings pane.

rupl (approved):
I did one final check and noticed that the primary email address was printed twice, so I swapped it out for a loop and displayed the email confirmation status similar to HID Contacts. With that, we can get this on dev and have others look at it :shipit:

PR #153, opened 2020-10-21
OPS-6803: fix inconsistent log conventions

rupl:
# OPS-6803

We had a couple issues reported in the ops ticket:

- `hid.user` had data of type `ObjectID` instead of strings sometimes
- `hid.error` had data of type `String` or sometimes `Error`  — but this seems to have been taken care of during 2020-09-30 deploy.

The PR has three fixes:
- attempts to get all the user ID stuff straightened out
- adds some additional attributes (email/ID) to a few logs
- fixes OAuth-authenticated requests to log the `client_id` inside `hid.oauth` instead of `hid.user`, and populated `hid.user` with correct data

## Testing

Can be complicated, but what I do is use GHO to log in on my local HID API:

- Add a GHO OAuth client to your local HID using the config from within GHO local drupal container.
- As drupal admin, go to https://gho.test/admin/config/social-api/social-auth/hid
- Under Advanced Settings, set the URL to be your local API: `http://api.hid.vm:3000`
- Open a shell to your GHO drupal container and run: `/sbin/ip route|awk '/default/ { print $3 }'` to get an IP to the host machine
- edit GHO drupal container's `/etc/hosts` to have `api.hid.vm` correspond to that IP
- tail logs in HID API container by running `docker-compose exec dev tail -f /var/log/local.log`
- in an incognito window, initiate an HID login from GHO website
- when prompted to enter a user/pass, choose something that exists, but has an expired password such as `verity@un.org`
- you should see the log entry contains a top-level property `user: { id: '[some string]' }` instead of the weird ObjectID which looks like this:

```json
"user":{"id":{"_bsontype":"ObjectID","id":{"0":91,"1":177,"2":216,"3":42,"4":43,"5":155,"6":251,"7":19,"8":17,"9":222,"10":191,"11":27}}}
```

- now login using a non-expired password account, and look for a log with the message `Successful authentication through OAuth token`
  - check that the `oauth: { client_id: 'something' }` exists
  - check that `user: { admin: boolean, id: 'string', email: 'address' }` exist

### DIFF
--- a/api/controllers/AuthController.js
+++ b/api/controllers/AuthController.js
@@ -88,7 +88,7 @@ async function loginHelper(request) {
         security: true,
         fail: true,
         user: {
-          id: user._id,
+          id: user.id,
           email,
         },
       },
@@ -103,7 +103,7 @@ async function loginHelper(request) {
         security: true,
         fail: true,
         user: {
-          id: user._id,
+          id: user.id,
           email,
         },
       },
@@ -119,6 +119,7 @@ async function loginHelper(request) {
         security: true,
         fail: true,
         user: {
+          id: user.id,
           email,
         },
       },
@@ -679,7 +680,7 @@ module.exports = {
             request,
             security: true,
             user: {
-              id: user._id,
+              id: user.id,
               email: user.email,
             },
             oauth: {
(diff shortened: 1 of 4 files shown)


rupl:
PS we can also just merge this to dev and check it in kibana

cafuego (approved):
I am 💯 in favour of YOYOing this to dev and then checking kibana.

rupl:
YOLO it is!

cafuego:
Wait, you went with LIVE not YEET ?

PR #158, opened 2020-10-29
[OPS-7052] bump node version

teodorescuserban:


rupl (approved):
I tried testing some API methods and the OAuth process via a local drupal site. Everything seems to work well and I didn't see any deprecation notices or other uncaught errors.

PR #159, opened 2020-11-03
HID-1968 oauth clients

left23:
1. Currently lists authorized clients
2. The "remove client" button is in place but not working
3. I've copied some functions from HID_v3 as placeholders for function logic. Not sure how much is applicable.
4. I've added the cd-button CSS for styles. I think these can be consolidated better nearer the end, once we know which components we will include.

Screenshot from 2020-11-03 14-00-03


rupl (approved):
@left23 we have some updated copy from Marina in the ticket, but otherwise this LGTM :shipit:

PR #160, opened 2020-11-03
HID-2126 Password reset form and 2FA activation

left23:
Branched from https://github.com/UN-OCHA/hid_api/pull/159

1. Currently placeholder markup only for both Password reset form, and 2FA tab
2. TOTP conditional based on vue app, but missing the "Activation Steps. TBD

Screenshot from 2020-11-03 16-46-03

Screenshot from 2020-11-03 16-46-11

Screenshot from 2020-11-03 16-49-00

### DIFF
--- a/templates/settings.html
+++ b/templates/settings.html
@@ -56,8 +56,7 @@
               </section>
 
               <section id="section-delete">
-                <h2>Delete Account</h2>
-                <p>Button here</p>
+                <%- include('delete'); %>
               </section>
             </div>
 
--- a/templates/settings.html
+++ b/templates/settings.html
@@ -12,26 +12,73 @@
             <div class="tabbed">
               <ul>
                 <li>
-                  <a href="#section-password">Password</a>
+                  <a href="#section-password" translate>Password</a>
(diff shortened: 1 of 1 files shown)


left23:
@rupl This is ready for review, with placeholder markup. 
I've added some placeholder JS, from the vue app, partly to preserve the logic. I can remove non-functional JS if you prefer, since it's in the other repo.

rupl:
I imagine we'll use the JS and the API to accomplish this instead of totally reinventing the wheel. Thanks for the starter and we'll get it functional once we have time.

left23:
@rupl Is this lack of functionality blocking merging this to dev?

rupl:
Maybe we can comment out any tab that isn't functional yet? I understand the need to not keep a bazillion parallel branches open, but we also don't want to provide a confusing UX by presenting buttons and tabs that do nothing.

left23:
@rupl This branch has only the Authorised App list tab now. Everything else is commented out.

left23:
Merging to dev.

PR #162, opened 2020-11-10
HID-1970 Delete account

left23:
1. add confirm message to button click
2. add totp-related js from vue app, commented out
3. add variant button style

### DIFF
--- a/templates/settings.html
+++ b/templates/settings.html
@@ -64,7 +64,9 @@
 
               <section id="section-delete">
                 <h2>Delete Account</h2>
-                <p>Button here</p>
+                <button type="button" class="cd-button cd-button--bold cd-button--delete" onclick="deleteAccount();">
+                  Delete
+                </button>
               </section>
             </div>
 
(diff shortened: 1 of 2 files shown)


rupl:
For this PR, I'm happy to also hide the tab and we'll unhide it in the same PR as we add code to make it functional and fully-tested.

templates/settings.html
```diff
@@ -192,4 +194,71 @@ <h2>Delete Account</h2>
     }
   }
 
+
+  async function deleteAccount(totpToken) {
```
rupl, line 198: Let's comment the JS out completely because `async` is not a thing that IE11 can handle, and will throw parse errors if we leave this available for the code to be executed.

templates/settings.html
```diff
@@ -192,4 +194,71 @@ <h2>Delete Account</h2>
...
+    //     .finally(() => {
+    //       this.pending = false;
+    //     });
+    // }
+  }
+
+  function resetTotpPrompt() {
+    this.needsTotp = false;
+    this.alreadyConfirmed = false;
+  }
```
rupl, line 262: You can also comment this out because the JS keyword `this` does not mean what it seems to mean, since we are no longer running the vode inside a Vue component.

templates/settings.html
rupl: Or sorry, I see what happened. Remove the word `async` and the rest can be retained. I'll strike out my comment above.

left23:
Closing without merging. I've merged with https://github.com/UN-OCHA/hid_api/pull/160 instead

PR #168, opened 2020-12-30
HID-2131: auth password v3

rupl:
# HID-2131 / HID-2144 / HID-2145

Ok I admit this PR has scope creep, but it all had to do with password reset or registration.

- Rename `isStrongPasswordV3` to `isStrongPassword` and delete the old v2 version using that name
- Remove all the temporary business logic that checked the request URL to determine which to use
- Ensure adoption is working within HID Auth
- Frontend/text updates to HID Auth
- Used HTML5 form validation on `new_password` and `register` forms.
- Consolidated JS form validation to one include file and used on both forms.
- a few lines of CSS to support the HTML5 validation. Also within the same include file as previous bullet.
- adopted same layout on most forms
- Fixed an HTML bug on the "message" template that was breaking CD.
- Logs which user failed to reset password in ELK.
- HID-2145: Errors during registration redirect back to registration form
- HID-2144: enforce passwords having to match on server during registration

### DIFF
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -2198,6 +2198,9 @@ module.exports = {
             request,
             security: true,
             fail: true,
+            user: {
+              id: request.payload.id,
+            },
           },
         );
         throw Boom.badRequest('Could not reset password');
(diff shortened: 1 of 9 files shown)


left23 (approved):
@rupl I've looked at this locally and it seems fine to me. If there's anything specific I should test, please let me know. Approving.

PR #161, opened 2020-11-04
HID-2126 Tailwind CSS framework and Cube CSS

left23:
Branches off #160 which has form elements to test Tailwind css classes

1. Tailwind styles to all pages and settings tabs.
2. Cube CSS for the flow, to control vertical rhythm on pages and within the form elements
3. disabled all previous framework styles
4. cd-button component for the buttons
5. commented out the settings tabs as in the dev branch so only functional code is displayed

**Tailwind**

I've installed the library based on https://tailwindcss.com/docs/installation
And based on looking at all the examples https://github.com/tailwindlabs/tailwindcss-setup-examples
and googling, the css is being generated and purged as part of the start script. 

We load the whole library for development (~ 2.3mB) and then rely on purge to remove most rules when we generate the tailwind-styles.css file for production (in my example, the CSS has only a handful of rules and is 14kB) https://tailwindcss.com/docs/controlling-file-size

I've basic configuration in place
https://tailwindcss.com/docs/configuration

https://tailwindcss.com/docs/adding-base-styles
I've done a simple example of extending the rules in `tailwind-source.css`, for the headings.
The preflight https://tailwindcss.com/docs/preflight unstyles headings https://tailwindcss.com/docs/preflight#headings-are-unstyled 

Screenshot from 2020-11-04 14-38-52

PR #170, opened 2021-01-06
HID-2147: frontend pipeline

rupl:
# HID-2147 / HID-2119 / OPS-7176

This branch reorganizes the frontend files and has some lite docs to define a workflow for styling individual pages. I took a chunk of your work in #161 and implemented it here. If I left out something major, we can ticket and return individually.

As is customary for this work, there is stuff you might consider scope creep. However in this case I was trying to rid us of cruft found while shuffling things around: bootstrap classes, layout inconsistencies, updates to make the new stack work smoothly, and small UX problems that I didn't want to ticket and return to later.

Happy to talk over this on a call, and also help you get set up on the new stack which I think will reduce some testing headaches.

## Testing

I'm looking for your input on the docs, and also general QA like:

- Can I still register (need to make a temp change in API to get around ReCAPTCHA)
- Can I reset a password?
- Can I login?


rupl:
@left23 ready for a final review. Our convo yesterday served as the majority of it, but if I missed any follow-ups please call them out. Ideally I'd like to close both this PR and #161.

left23:
@rupl Looks good to me. I've tweaked a few things that I think are uncontroversial but feel free to reverse or improve as you see fit.

1. I noticed there is no width restriction to the form elements yet - did you have specific plans for this?
2. The OCHA services button font size should be 14px. The cd-button component makes it 16px - I've added the font-size to the existing `.ocha-dropdown__see-all` rule for now, and I think I should set this in the cd.css in the common_design to avoid this issue in the future
3. The /register and /password buttons needed the cd-button classes added - I've removed `btn-primary` and replaced with classes matching those on main login form button. *There may be other forms and submit buttons that need similar treatment, depending on desired styles.
4. I added some colour rules from https://github.com/UN-OCHA/hid_api/pull/161/files#diff-1671da1302cd5c094dd762e76e142d8c1caf8f276eb884b86ef1630507e81ed2 so the H1, h2 and links are HID brand colours (as before)

I think it's ok to close #161 now. Everything of import seems to be included in this branch now.

left23 (approved):

left23:
I'm able to login and reset passwords as usual. I had issues registering a new account `There was a problem validating your registration. Please try again.` but I get the impression from the PR description that is expected.

left23:
@rupl One more thing - I'm not sure the font Exo made it into the latest styles. I can't see a Fonts directory anymore so maybe this is intentional.
It was part of main.css which is gone now https://github.com/UN-OCHA/hid_api/pull/170/files#diff-b15932692c619f9a334ecb156ac167557c3af214a8a3df8990c09232ac8a5baaL66

rupl:
@left regarding my comment about reCAPTCHA, you can comment out the validation and return `true` so that registrations finish with success: https://github.com/UN-OCHA/hid_api/blob/dev/api/controllers/ViewController.js#L184

rupl:
@left23 as for Exo, I searched once more and couldn't find it being referenced in our CSS files. Did I miss a reference? I did intend to remove it completely.

left23:
> @left23 as for Exo, I searched once more and couldn't find it being referenced in our CSS files. Did I miss a reference? I did intend to remove it completely.

Nope, it's gone. I just wanted to double check that was the intention. Thorough :)

left23:
@rupl I've confirmed registration works too.

rupl:
Nice! Then let's give this a try on dev.

PR #171, opened 2021-01-12
HID-1965: HID Auth profile management

rupl:
# HID-1965

This branch introduces the ability for Auth users to do basic profile management. Specifically:

- edit first name
- edit last name
- choose primary email address (must be confirmed)
- add a new email address (confirmation process unchanged)

Things you can't do yet:

- **delete email addresses:** All our UI patterns for this were using client-side JS and API calls. I didn't want to bite that bullet yet which would have required installing a bunch of JS libs/polyfills. Additionally, I couldn't think of a simple method of offering delete buttons to the non-primary email addresses using only HTML forms. I'm all ears.

Additionally, some follow-ups to #160 and #170 were added to this branch as they were blocking progress:

- CSS Vars ponyfill for IE11 
- I removed the old band-aid JS we had on HID because the CD Header is fully standardized now.

## Server-side adjustments

On the server, a few adjustments were made to allow internal use of existing API calls. Whenever a function was used, I added a second `internalArgs` object which will never be supplied by a "normal" API call. When the function sees that `internalArgs` was supplied, it will opt to pull the parameters and payload from that object.

This convention is documented in `CONTRIBUTING.md`

## Testing

You should be able to use either your personal account, or register a new one (remember to fake-out the reCAPTCHA if you register anew). Once logged in, use the CD user menu (which now says your name!)

Remember to test in IE! I did my best but another pair of 👀  is very much appreciated. Just FYI: my browser sessions were not persisting in browserstack using Win7 + IE11, and I ended up testing in Win10 + IE11.

## Other tickets

There are minor fixes for the following other tickets:

- HID-1878
- HID-2147
- HID-2150
- OPS-7176


api/controllers/ViewController.js
```diff
@@ -455,19 +458,180 @@ module.exports = {
...
+        await UserController.addEmail({}, {
+          userId: cookie.userId,
+          email: request.payload.email_new,
+          appValidationUrl: _buildRequestUrl(request, 'verify2'),
+        }).then(data => {
+          cookie.alert.message += `<p>A confirmation email has been sent to ${request.payload.email_new}.</p>`;
+        }).catch(err => {
+          // Read our error and show some user feedback.
+          if (err.message.indexOf('Email is not unique') !== -1) {
+            cookie.alert.message += `<p>The address ${request.payload.email_new} is already added to your account.</p>`;
```
rupl, line 622: @left23 per Flowdock discussion, this is where the error gets printed. but if we look deeper for logs that have a `[UserController->addEmail]` prefix, that would probably reveal the true cause for error.

rupl: Yes it seems the `UserController.addEmail()` error doesn't differentiate between your account or someone else's. A simple change of wording when we handle the error will clear things up IMO.

> That email address has already been registered.

left23 (approved):
Looks great @rupl Approving.
Testing in IE and other browsers was all good, and general functionality testing was fine too.

I pointed out the generic error message when I tried to add an email that already existed for a second account to my current account. I understand this is being dealt with in a following PR.

I made a few commits for adding the cd-button classes to the "Reset password" and "Register" buttons for consistency. And changed the "Want to logout?" link to look like a button reading "Logout".

rupl:
Nice, thanks for the review. I've addressed the puzzling error in the next branch so I'll merge this onto dev.

PR #172, opened 2021-01-13
HID-2153: manage emails on profile

rupl:
# HID-2153

This branch finishes the profile management section by allowing all aspects of emails to be managed from one screen:

- Add a new email to your profile
- Re-send an email confirmation to unconfirmed addresses
- Delete an email from your profile

I tried to be very thorough in validating form submissions even to the point of editing the DOM and getting around the basic client-side validations. If you can find a way to break it I will be very interested to see what you come up with.

- Trying to set an email as primary with the radio button, then click <kbd>DELETE</kbd> on that same email address
- Edit DOM to send confirmations to confirmed emails
- Edit DOM to delete primary address
- Edit DOM to set an unconfirmed address as primary

I had to tweak the internals of several functions but tested them along the way. I retested each individual endpoint, in addition to all the ones which require multiple steps (e.g. firing API call to generate confirmation email => clicking email). Some other tweaks were done to support the UX of our profile editing screens.

- The error discovered in #171 about "this email is already on your account" has been addressed — we differentiate between ones on your profile, and when another HID user has claimed an email. No new info is surfaced by doing this. The API was already outputting this info.
- If a user session is detected when confirmation links are clicked, it no longer forces you to log in again. It redirects to your profile so that people clicking <kbd>CONFIRM</kbd> then returning via email will see the same page.

I added a few more buttons to the user landing page so all their options are spelled out for them.

I did test in Safari, Chrome, FF, and IE11. Happy to have another pair of 👀 regardless.

### DIFF
--- a/assets/css/page.css
+++ b/assets/css/page.css
@@ -203,7 +203,7 @@ input:focus:invalid {
   outline-offset: 0;
 }
 
-.form-field code {
+form code {
   display: inline-block;
   background: #eee;
   border-radius: 3px;
--- a/assets/css/page.css
+++ b/assets/css/page.css
@@ -81,6 +81,14 @@ h1, h2, h3, h4, h5 {
   background: #007faa;
 }
 
+.cd-button--danger {
+  background-color: var(--cd-highlight-red);
+}
+
+.cd-button[disabled] {
+  opacity: .666;
+}
+
 
 /**
  * Global layout
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -1551,6 +1551,7 @@ module.exports = {
       appValidationUrl = request.payload.app_validation_url;
     }
 
+    // Is the payload complete enough to take action?
     if (!appValidationUrl || !email) {
       logger.warn(
         '[UserController->addEmail] Either email or app_validation_url was not provided',
@@ -1565,6 +1566,7 @@ module.exports = {
       throw Boom.badRequest('Required parameters not present in payload');
     }
 
+    // Is the verification link pointing to a domain in our allow-list?
     if (!HelperService.isAuthorizedUrl(appValidationUrl)) {
       logger.warn(
         `[UserController->addEmail] app_validation_url ${appValidationUrl} is not in authorizedDomains allowlist`,
@@ -1577,18 +1579,7 @@ module.exports = {
       throw Boom.badRequest('Invalid app_validation_url');
     }
 
-    // Make sure email added is unique
-    const erecord = await User.findOne({ 'emails.email': email });
-    if (erecord) {
-      logger.warn(
-        `[UserController->addEmail] Email ${email} is not unique`,
-        {
-          request,
-          fail: true,
-        },
-      );
-      throw Boom.badRequest('Email is not unique');
-    }
+    // Does the target user exist?
     const record = await User.findOne({ _id: userId });
     if (!record) {
       logger.warn(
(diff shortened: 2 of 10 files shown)


assets/css/page.css
```diff
@@ -81,6 +81,14 @@ h1, h2, h3, h4, h5 {
...
 }
 
+.cd-button--danger {
+  background-color: var(--cd-highlight-red);
+}
+
+.cd-button[disabled] {
+  opacity: .666;
+}
+
```
left23, line 91: These will be good to add to the cd-button component

left23:
This looks great. I like the UI additions.

I can confirm the error message is clear when I tried to add an existing registered email.

My first action was to attempt to delete the secondary email while also selecting its radio button, and you'd predicted that nicely with an informative error message :)

Initially the message about the `[confirm]` button confused me - I thought it was referring to the `Update email` button that was mislabeled because I didn't have an email pending confirmation so I saw no `confirm` button. However once I went through the steps it became clear. Maybe depending on what other people give as feedback, it's a possibility to print this message conditionally, only when an email is pending confirmation?

As you suggested, the "Primary" label and "Delete" button might cause confusion because of the similar style but different function. However I'd be curious if comes up when others are testing. The only thing I can suggest is we come up with a style for labels that distinguishes better (although not sure what the visual style would be at the moment) or maybe a different placement in relation to the email (label to come before the email, button after?) Maybe this becomes more clear when work on the 2FA Settings is underway, as I think there is a status label there as well. There is also the Profile display with the status labels that could be made consistent
Screenshot from 2021-01-14 11-29-34

I was able to submit the "Manage email address" form with an incomplete email address. I got a "Email settings saved" confirmation but it didn't add the email address (which is good)
Screenshot from 2021-01-14 10-46-57
Screenshot from 2021-01-14 10-47-05

I tried to register with a similar incomplete email address and I get an error (not specific to an invalid email but enough so the form doesn't submit)
Screenshot from 2021-01-14 11-11-08

I don't have any concrete suggestions here. After reading a bit about using type=email versus patterns it seems what you have is most appropriate for validation but maybe you have ideas. I'm also happy to entertain some of the burden being on the user but maybe this is not the approach you had hoped to take.

I'll do some cross-browser testing now.

left23:
@rupl Browser testing was good. I noticed a display issue on narrow viewports when both the Delete and Confirm button are present. I've committed a small CSS adjustment to prevent a horizontal scroll in those cases.
I tried to align the second button right but it didn't seem worth it.
End result:
Screenshot from 2021-01-14 12-16-55
Screenshot from 2021-01-14 12-17-08

left23:
> 
    Trying to set an email as primary with the radio button, then click DELETE on that same email address
    Edit DOM to send confirmations to confirmed emails
    Edit DOM to delete primary address
    Edit DOM to set an unconfirmed address as primary


1. Did so without prompting, see my first comment

2. Fails as desired with nice error message
Screenshot from 2021-01-14 13-05-52

3. Fails as desired with nice error message
Screenshot from 2021-01-14 13-09-35

4. Fails as desired with nice error message
Screenshot from 2021-01-14 13-12-33

left23:
@rupl We'd briefly discussed the 2 cancel buttons and whether improvements can be made here. I think the two form actions feel separate enough that I didn't notice nor was I confused, so while it might be a bit redundant, I think it's adequate to leave as is.

rupl:
Thanks for the thorough review! Interesting about the mobile display bug. I had personally tried to alter the layout in a way that kept everything on one line, but space is admittedly limited when both Delete+Confirm are on the same row as an email address, and of course being able to read/press everything is more important than the aesthetics of one-line-per-email.

rupl:
Regarding email addresses. I suppose matching client-side logic with the server-side will be in the best interest of the user. I'll file a ticket to follow-up so that the regex can be thoroughly tested and compared to the server-side. Hopefully using the server-side regex will work in all the browsers. 🤞

PR #173, opened 2021-01-14
HID-2154 profile cards

left23:
Details on [HID-2154](https://humanitarian.atlassian.net/browse/HID-2154)

There may be more work needed to make this consistent with the Profile Edit page. And keeping, ditching or reducing the cd-grid component is up for discussion (based on whether it's useful elsewhere)

### DIFF
--- a/assets/css/page.css
+++ b/assets/css/page.css
@@ -28,6 +28,9 @@
 
   /* darken */
   --cd-primary-color--dark: hsl(var(--cd-primary-color-h), var(--cd-primary-color-s), calc(var(--cd-primary-color-l) - 26%));
+
+  /* CD colours for components */
+  --cd-blue-grey: #e6ecf1;
 }
 
 /**
--- a/assets/css/cd-card.css
+++ b/assets/css/cd-card.css
@@ -1,0 +1,64 @@
+.cd-card {
+  flex-direction: column;
+  /* Override grid background color from grid */
+  background-color: var(--cd-white);
+  border: 2px solid var(--cd-blue-grey);
+}
+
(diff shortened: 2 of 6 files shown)


assets/css/page.css
```diff
@@ -28,6 +28,9 @@
 
   /* darken */
   --cd-primary-color--dark: hsl(var(--cd-primary-color-h), var(--cd-primary-color-s), calc(var(--cd-primary-color-l) - 26%));
+
+  /* CD colours for components */
+  --cd-blue-grey: #e6ecf1;
```
left23, line 33: This is a bit silly but because we are working on CD v2 https://github.com/UN-OCHA/common_design/tree/cd-226-design-system, this colour var is in that branch which is why I needed to add it explicitly here.
We're using CD v1 as HID's cd.css so some things might be missing to do with specific components, as the components are being developed further in v2.
This is just a heads up. We might encounter more annoying issues with this later on. If we decided to use v2 as cd.css for HID, we would end up with the Full width header and footer (which might be desirable) along with more CSS custom properties.

rupl: Oh, interesting. I hadn't considered pulling the CD artifact from the DS branch. In your opinion, should we brave it? I'm happy to try it, especially if it makes adopting a few more components easier.

left23: Let me investigate this a bit more. I think it depends on timing mainly. Whether the v2 is "ready enough" to coincide with hid_api being near completion. Also I'm not entirely sure what plans we have for unrolling the CD v2 in general. To be continued.

PR #174, opened 2021-01-14
HID-2009: manage oauth clients on user profile

rupl:
# HID-2009

This PR allows user's logged into HID Auth to list and revoke OAuth Clients they've authorized for their account.

## New API Endpoint

There was no method to do this internally, and the old system was relying on a JS client to manually modify the user object, then `PUT` the entire user back into the DB.

I didn't like that options, so to facilitate this action internally I created a new API endpoint:

- `DELETE /api/v3/user/{id}/clients/{client}`
- the handler is `UserController.revokeOauthClient`

@orakili if you could review this endpoint and try to break it I would very much appreciate it. There are new docs available at http://api.hid.vm/docs/v3/#/user/delete_user__id__clients__client_ when you're running the API locally.

## HID Auth UX + CD components

@left23 the remaining part is again the UX and more adoption of CD components. I've implemented another `cd-button` for the revoke button and moved the buttons all to be left of the label so that they're lined up with one another.

I also implemented `cd-alert` styles from your feature branch, but did not change much markup. I do intend to change it, but altering the markup all over the website felt like scope creep once I started, so I went with CSS-only changes. Despite its incompleteness, I am loving how it looks compared to the old styles even without all parts of `cd-alert`, namely the missing icons.

For testing the same as profile editing applies, feel free to tweak DOM and try to break it. Tell me if any of the errors sound weird. Some of them are a little vague, but I thought group discussion would improvement quicker than me sitting alone in my corner.

I also got rid of old client-side code, and moved the tabbing JS into its own file that we might easily re-use it. Eventually I intend to offer each settings tab at its own URL, and then convert the tabs to plain hyperlinks and eventually remove the need for tab JS.

### DIFF
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -2565,6 +2565,17 @@ module.exports = {
    * tags:
    *   - user
    * summary: Revokes one OAuth Client from a user's profile.
+   * parameters:
+   *   - name: id
+   *     in: path
+   *     description: The user ID
+   *     required: true
+   *     type: string
+   *   - name: client
+   *     in: path
+   *     description: The OAuth Client ID
+   *     required: true
+   *     type: string
    * responses:
    *   '200':
    *     description: OAuth Client revoked successfully. Returns user object.
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -2687,6 +2687,8 @@ module.exports = {
           stack_trace: err.stack,
         },
       );
+
+      throw Boom.internal('Internal server error.');
     }
   },
 };
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -2559,4 +2559,123 @@ module.exports = {
     );
     return user;
   },
+
+  /*
+   * @api [delete] /user/{id}/clients/{client}
+   * tags:
+   *   - user
+   * summary: Revokes one OAuth Client from a user's profile.
+   * responses:
+   *   '200':
+   *     description: OAuth Client revoked successfully. Returns user object.
+   *   '400':
+   *     description: Bad request. See response body for details.
+   *   '401':
+   *     description: Requesting user lacks permission to view requested user.
+   *   '404':
+   *     description: Requested user not found.
+   */
+  async revokeOauthClient(request, internalArgs) {
+    let userId, clientId;
+
+    if (internalArgs && internalArgs.userId && internalArgs.clientId) {
+      userId = internalArgs.userId;
+      clientId = internalArgs.clientId;
+    } else {
+      userId = request.params.id;
+      clientId = request.params.client;
+    }
+
+    // Validate presence of userId param.
+    if (!userId) {
+      logger.warn(
+        '[UserController->revokeOauthClient] No userId provided',
+        {
+          request,
+          fail: true,
+        },
+      );
+      throw Boom.badRequest('No userId provided.');
+    }
+
+    // Validate presence of clientId param.
+    if (!clientId) {
+      logger.warn(
+        '[UserController->revokeOauthClient] No clientId provided',
+        {
+          request,
+          fail: true,
+          user: {
+            id: userId,
+          },
+        },
+      );
+      throw Boom.badRequest('No clientId provided.');
+    }
+
+    // Look up user from DB.
+    const user = await User.findOne({ _id: userId });
+
+    // Validate that user was found in DB.
+    if (!user) {
+      logger.warn(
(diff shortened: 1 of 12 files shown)


left23:
@rupl On first inspection, this looks great.
I've made commits based on https://github.com/UN-OCHA/common_design/pull/107

rupl:
@left23 thanks! And reviewing your changes made me notice that I wrote "Delete [client]" as the `aria-label` when we're actually revoking clients.

left23 (approved):
@rupl Approving this. Looks good and works well. I tried to break it, could not :)

api/controllers/UserController.js
```diff
@@ -2559,4 +2559,123 @@ module.exports = {
     );
     return user;
   },
+
+  /*
+   * @api [delete] /user/{id}/clients/{client}
+   * tags:
+   *   - user
+   * summary: Revokes one OAuth Client from a user's profile.
+   * responses:
```
orakili, line 2568: Wouldn't be better to add the `id` and `client` parameters here to match lines  2585 and 2586? That would also help when trying the endpoint in swagger.

api/controllers/ViewController.js
```diff
@@ -800,9 +800,92 @@ module.exports = {
...
+    }
+
+    // Load current user from DB.
+    const user = await User.findOne({ _id: cookie.userId });
+
+    // Set up user feedback.
+    let reasons = [];
+    let alert = {};
+
+    console.log('🐛', request.payload);
```
orakili, line 839: Is that supposed to be there?

api/controllers/UserController.js
```diff
@@ -2559,4 +2559,123 @@ module.exports = {
...
+      return user;
+    } catch (err) {
+      logger.error(
+        `[UserController->revokeOauthClient] ${err.message}`,
+        {
+          fail: true,
+          stack_trace: err.stack,
+        },
+      );
+    }
```
orakili, line 2679: No `Boom` here. Is that fine? Wouldn't that result in no feedback to the user on line 880 of `api/controllers/ViewController.js`?

docs/v3/hid.yml
```diff
@@ -907,4 +907,18 @@ paths:
         '400':
           description: Bad request. See response body for details.
       security: []
+  '/user/{id}/clients/{client}':
+    delete:
+      tags:
+        - user
+      summary: Revokes one OAuth Client from a user's profile.
+      responses:
```
orakili, line 915: See comment above about `id` and `client` parameters.

templates/footer.ejs
```diff
@@ -25,7 +25,7 @@
         <div class="cd-footer__section cd-footer__section--social">
           <a href="https://github.com/UN-OCHA" class="cd-footer-social__link">
             <span class="element-invisible">Github</span>
-            <svg class="cd-icon cd-icon--github">
+            <svg class="cd-icon cd-icon--github" width="32" height="32" aria-hidden="true" focusable="false">
               <use xlink:href="#cd-icon--sm-gh-def"></use>
             </svg>
             <svg class="cd-icon cd-icon--github hover-style">
```
orakili, line 31: Shouldn't this also have the `aria-hidden="true" focusable="false"` attributes?

orakili (approved):
Looks good and didn't find any issues during local testing regarding the core features. See comments below for a few things to check before merging.

api/controllers/ViewController.js
rupl: whoops, no never. the 🐛 is supposed to be my easy way to spot and remove debug stuff 🙃

api/controllers/UserController.js
rupl: Great idea! I never quite understood that we can document params in the path, but it was [easy to find out how](https://swagger.io/docs/specification/describing-parameters/) and allows the API call to be tested in our Swagger site.

docs/v3/hid.yml
rupl: This is an auto-generated file assembled from the inline code blocks by running `doocker-compose exec api npm run docs`

api/controllers/UserController.js
rupl: I suppose I never tested that particular code path because all the previous validations prevented me from getting there. I'll add a `throw Boom.internal()` in case there are issues with DB connection or something.

rupl: I manually wrote `throw Error();` after `await user.save();` and this is what it showed me.

Screen Shot 2021-01-19 at 10 52 09

When we craft other more specific errors, we provide some instructions, but without "real-world" errors I'm not sure what to write, and that will at least be something for now.

PR #176, opened 2021-01-18
HID-2155 alerts

left23:
~~This branch merges off https://github.com/UN-OCHA/hid_api/pull/175~~

It consists of implementing the most recent cd-alert component
https://github.com/UN-OCHA/common_design/tree/cd-226-design-system/components/cd-alert

https://feature.commondesign-unocha-org.ahconu.org/demo#cd-alert

### DIFF
--- a/templates/alert.html
+++ b/templates/alert.html
@@ -1,3 +1,5 @@
+<style><% include ../assets/css/cd-alert.css %></style>
+
 <% if (typeof alert !== 'undefined' && alert.type && alert.message) { %>
   <div class="alert alert--inline alert--<%= alert.type %> [ flow ]" role="alert">
     <div class="alert__inner">
--- a/templates/alert.html
+++ b/templates/alert.html
@@ -1,8 +1,8 @@
 <style><% include ../assets/css/cd-alert.css %></style>
 
 <% if (typeof alert !== 'undefined' && alert.type && alert.message) { %>
-  <div class="alert alert--inline alert--<%= alert.type %> [ flow ]" role="alert">
-    <div class="alert__inner">
+  <div class="cd-alert cd-alert--inline cd-alert--<%= alert.type %> [ flow ]" role="alert">
+    <div class="cd-alert__inner">
       <%- alert.message %>
     </div>
   </div>
--- a/templates/alert.html
+++ b/templates/alert.html
@@ -1,8 +1,8 @@
 <style><% include ../assets/css/cd-alert.css %></style>
 
 <% if (typeof alert !== 'undefined' && alert.type && alert.message) { %>
-  <div class="cd-alert cd-alert--inline cd-alert--<%= alert.type %> [ flow ]" role="alert">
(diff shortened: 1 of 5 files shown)


templates/alert.html
```diff
@@ -1,6 +1,8 @@
+<style><% include ../assets/css/cd-alert.css %></style>
+
 <% if (typeof alert !== 'undefined' && alert.type && alert.message) { %>
-  <div class="alert alert--inline alert--<%= alert.type %> [ flow ]" role="alert">
-    <div class="alert__inner">
+  <div class="cd-alert cd-alert--inline cd-alert--<%= alert.type %> [ flow ]" role="alert">
```
rupl, line 4: we can totally remove `cd-alert--inline` — I'm not sure where/when it was being used, but it's officially cruft now.

templates/alert.html
```diff
@@ -1,6 +1,8 @@
...
+    <div class="cd-alert__inner">
```
rupl, line 5: Same for the `.cd-alert__inner` — if your reference markup isn't using it (and I son't see the class in the official component) then we can cut it from HID.

rupl (approved):
Looks great and I re-tested all the forms on the site. Things seem to be working as expected 👍 

I noticed a side-effect of adopting the real component is the spacing of many errors changed. Since consistency across sites is key, I'll try to collect some examples I find for discussion, so that HID can better match the Drupal family.

I think the main point of discussion is: **do we try and recommend markup conventions inside the alert?** HID has `<p>` and `<ul>` tags in its errors (plus that one `<input>`). Of course, the easiest solution is for HID to do what Drupal doe, so that will be mostly a Drupal survey so that we can mimic in HID.

templates/alert.html
rupl: I popped a commit onto the branch

templates/alert.html
rupl: I attempted to remove this wrapper element, but it does seem to be necessary in order to allow the yet-to-be-added icon to appear in the correct location. When I removed the container, various block-level tags started displaying in a row and changing the class to be `cd-alert__message` seems like the easy change here, so that it both means something, and leaves room for the icon to be added.

left23:
> spacing

I noticed the extra markup. I'd only planned on a heading and a short bit of text in the alerts and I can see that isn't sufficient. 
I think it makes sense to add some rules for `<p>` and `<ul>` as part of the cd-alert component, like https://github.com/UN-OCHA/hid_api/blob/dcdbf9fc9ec6db9172f3f409430e792c7b711819/assets/css/page.css#L143

Drupal outputs errors as <li> items in a <ul> so the cd-alert doesn't really cater for that either at the moment.

One thing I noticed is more missing variables:
```
--cd-blue--bright: #80cbff;
--cd-grey--light: #f2f2f2;
--cd-grey--mid: #595959;
--cd-grey--dark: #4a4a4a;
```
These should be added to the global CSS

PR #177, opened 2021-01-20
HID-1970: delete account

rupl:
# HID-2164

@orakili our TOTP logic for backup codes needed some love. For one, when setting up 2FA via API calls, it's possible to skip the `POST /totp/codes` API call and end up with half-configured 2FA that our `User.backupCodeIndex()` function was choking on. I added a quick guard to check each property in the nested objects to avoid such a heartbreak 💔 

Additionally, backup codes weren't getting removed — even when we logged that they were 🤦 — but that should be cleaned up now as well. To test the halfway 2FA state, run the following calls:

- `POST /jsonwebtoken` (to get your API key)
- `POST /totp/config` it will output a base64 image but also a plaintext config string that you can paste to an Auth app
- `POST /totp` — enable 2FA once your Auth app is generating codes
- Now you're at the halfway point. Try doing something that requires 2FA (login is fine) and when prompted for a token, enter  anything except 6 integers to trigger the backup code lookup. It would report that you have an `Invalid backup code !`

To test backup code deletion, first generate codes:

- `POST /totp/codes` — generate backup codes (see HID-2075 for lengthy thoughts on this API call)
- Take another action requiring 2FA. This time, use a backup code.
- Do one more action requiring 2FA. Use the **same backup code** and confirm that it prevents the action
- If you can think of another workflow and try to bust it, I'll be glad to have your feedback.

# HID-1970

@left23 non-admin users can delete their own account from settings. 2FA users require successful TOTP challenge.

- Test an admin user (ping me if you need assistance setting up an admin)
- Test a normal user without 2FA enabled (fresh registrations will be just fine)
- Test a normal user with 2FA enabled (again I'm here to help you set this up if the above instructions aren't enough)

Since you'll probably want to register a dummy account to repeatedly test this, I direct you toward the ReCAPTCHA which you can comment out:

https://github.com/UN-OCHA/hid_api/blob/bc7f93d7867e516bb7e2ab248615683b16fc9c1d/api/controllers/ViewController.js#L188

Like the other forms, I encourage you to try and trick the system any way you can: edit the DOM to change form validation, paste the form from a non-admin into the Admin's DOM, change primary email after loading form, you name it. I tried to be extremely thorough.

### DIFF
--- a/api/models/User.js
+++ b/api/models/User.js
@@ -1317,7 +1317,7 @@ UserSchema.methods = {
 
   backupCodeIndex(code) {
     let index = -1;
-    let numCodes = this.totpConf && this.totpConf.backupCodes && this.totpConf.backupCodes.length;
+    let numCodes = this.totpConf && this.totpConf.backupCodes ? this.totpConf.backupCodes.length : 0;
     for (let i = 0; i < numCodes; i++) {
       if (Bcrypt.compareSync(code, this.totpConf.backupCodes[i])) {
         index = i;
--- a/api/models/User.js
+++ b/api/models/User.js
@@ -1317,7 +1317,8 @@ UserSchema.methods = {
 
   backupCodeIndex(code) {
     let index = -1;
-    for (let i = 0; i < this.totpConf.backupCodes.length; i += 1) {
+    let numCodes = this.totpConf && this.totpConf.backupCodes && this.totpConf.backupCodes.length;
+    for (let i = 0; i < numCodes; i++) {
       if (Bcrypt.compareSync(code, this.totpConf.backupCodes[i])) {
         index = i;
       }
--- a/api/policies/AuthPolicy.js
+++ b/api/policies/AuthPolicy.js
@@ -56,8 +56,8 @@ async function isTOTPValid(user, token) {
   }
 
   // Remove backup code so it can't be reused.
-  user.totpConf.backupCodes.slice(index, 1);
-  user.markModified('totpConf');
+  user.totpConf.backupCodes.splice(index, 1);
+  user.markModified('totpConf.backupCodes');
   await user.save();
 
   logger.info(
--- a/api/policies/AuthPolicy.js
+++ b/api/policies/AuthPolicy.js
@@ -34,7 +34,10 @@ async function isTOTPValid(user, token) {
     }
     logger.warn(
       `[AuthPolicy->isTOTPValid] Invalid TOTP token ${token}`,
-      { security: true },
+      {
+        security: true,
+        fail: true,
+      },
     );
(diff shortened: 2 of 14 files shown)


api/models/User.js
```diff
@@ -1317,7 +1317,8 @@ UserSchema.methods = {
 
   backupCodeIndex(code) {
     let index = -1;
-    for (let i = 0; i < this.totpConf.backupCodes.length; i += 1) {
+    let numCodes = this.totpConf && this.totpConf.backupCodes && this.totpConf.backupCodes.length;
+    for (let i = 0; i < numCodes; i++) {
```
rupl, line 1321: @orakili upon looking at this after sleeping, I thought it good to call out the three lines of code you're auditing 😁  this is one spot.

api/policies/AuthPolicy.js
```diff
@@ -44,18 +47,30 @@ async function isTOTPValid(user, token) {
...
     );
     throw Boom.unauthorized('Invalid backup code !', 'totp');
   }
 
-  // remove backup code so it can't be reused
-  user.totpConf.backupCodes.slice(index, 1);
-  user.markModified('totpConf');
+  // Remove backup code so it can't be reused.
+  user.totpConf.backupCodes.splice(index, 1);
+  user.markModified('totpConf.backupCodes');
```
rupl, line 60: @orakili second block of code. most everything else is either for Lisa or logging tweaks

left23 (approved):
@rupl Approving.

Normal user without 2FA enabled works.
Screenshot from 2021-01-21 17-47-39

Test a normal user with 2FA enabled works.
Screenshot from 2021-01-21 17-47-39

The UI is clear (and looks nice!) and when entering emails incorrectly, the prompts are informative.

I wasn't able to test as admin.

orakili (approved):
Working well. I couldn't find a way to bust it. I'm just wondering if we want to inform the user that it doesn't have any backup codes left once all of them have been used.

api/models/User.js
```diff
@@ -1317,7 +1317,8 @@ UserSchema.methods = {
 
   backupCodeIndex(code) {
     let index = -1;
-    for (let i = 0; i < this.totpConf.backupCodes.length; i += 1) {
+    let numCodes = this.totpConf && this.totpConf.backupCodes && this.totpConf.backupCodes.length;
```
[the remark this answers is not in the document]
rupl: you're right, probably safest to assign to a number in all cases. Thanks!

As for notifying, I have ticketed it and put in backlog. Right now, we only report the success or failure of the TOTP challenge, and there's no external-facing log about backup codes being removed. We do distinguish between failed time-based token vs failed backup code, but I'm not even sure it's necessary to do that.

HID-2165

PR #178, opened 2021-01-21
HID-2159: manage 2FA settings

rupl:
# HID-2159

HID Auth users now have 2FA settings available. You should be able to:

- Enable 2FA and it shows up during login, for instance.
- Go back to settings and disable 2FA by again entering a token or backup code.

I updated the following methods but tested them pretty thoroughly and the changes were very small anyway:

- `TOTPController.generateConfig()`
- `TOTPController.generateBackupCodes()`
- `TOTPController.enable()`
- `TOTPController.disable()`

### DIFF
--- a/api/controllers/AuthController.js
+++ b/api/controllers/AuthController.js
@@ -415,7 +415,7 @@ module.exports = {
         return loginRedirect(request, reply);
       } catch (err) {
         const alert = {
-          type: 'danger',
+          type: 'error',
           message: err.output.payload.message,
         };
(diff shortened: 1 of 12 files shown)


left23 (approved):
@rupl Approving.

I can confirm enabling and disabling 2FA works as expected, and the pages looks great.
I got prompted when I logged in as expected with 2FA enabled.
Backup codes generated.
And when I enter incorrect codes the messages are clear.

PR #184, opened 2021-01-25
HID-2142: log when listing OAuth client(s)

rupl:
# HID-2142

Added logging to `ClientController.find()` and also corrected a few discrepancies with docs:

- the default `sort` was actually `_id` but we advertise `name` — it should really be `name` now 
- the default pagination was 50 but it's 100. Docs updated to say 100.
- the logs now list how many records were displayed, and who asked to see them in ELK meta fields
- the pre-existing 404 log also contains the security-related meta fields

Endpoints to test (you must have an HID user who has `user.is_admin=true`:
- `GET /api/v3/client`
- `GET /api/v3/client/{id}`

### DIFF
--- a/api/controllers/ClientController.js
+++ b/api/controllers/ClientController.js
@@ -165,6 +165,8 @@ module.exports = {
     const sort = request.query.sort || 'name';
     options.sort = sort;
     const results = await HelperService.find(Client, criteria, options);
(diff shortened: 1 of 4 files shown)


orakili (approved):
Working as advertised.

PR #185, opened 2021-01-26
HID-2117: fix ugly errors during Authorization process

rupl:
# HID-2117

Sometimes HID outputs ugly errors that make it look unstable. It turned out the errors were being displayed that way on purpose (more or less). So this PR fixes the branding issues and makes the error much more user-friendly.

## Testing

Unfortunately to test this you have to craft an OAuth request. I have a puppeteer script sitting around which lets me skip manually logging in and it autofills forms and stuff. I can put it in a repo if you want, but you can also trigger the Authorization by using the following URL template:

```
GET http://api.hid.vm/oauth/authorize
  ?response_type=token
  &client_id=chris-debug
  &client_secret=0123456789abcdefghijklmnopqrstuvwxyz
  &redirect_uri=http%3A%2F%2Fchris.debug.test%2Flogin%2Fcallback
  &state=12345
  &nonce=19207
```

You will have to create that client, or replace the dummy data shown here with a client from your DB. Generally I recommend against using Prod config because if all goes well, you end up at a prod URL which collects analytics.

 The three conditions affecting the PR are:
- FAIL `client_id` — send something that does NOT exist in the DB to trigger the error
- FAIL `redirect_uri` — send something that does NOT exist in the DB (either `redirect_uri` or in the array of `redirect_urls`)
- PASS: when you craft a valid URL, does it still work?

### DIFF
--- a/api/controllers/AuthController.js
+++ b/api/controllers/AuthController.js
@@ -554,9 +554,9 @@ module.exports = {
       const clientId = request.query.client_id;
       user.sanitize(user);
       request.auth.credentials = user;
-      const result = await oauth.authorize(request, reply, {}, async (clientID, redirect, done) => {
+      const result = await oauth.authorize(request, reply, {}, async (oauthClientId, redirect, done) => {
         try {
-          const client = await Client.findOne({ id: clientID });
+          const client = await Client.findOne({ id: oauthClientId });
           if (!client || !client.id) {
             logger.warn(
               '[AuthController->authorizeDialogOauth2] Unsuccessful OAuth2 authorization because client was not found',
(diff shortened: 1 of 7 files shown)


rupl:
Whoops I forgot the screenshots

### Before

HID-2117-old-error

### After

HID-2117-new-error

rupl:
@orakili for testing use https://github.com/UN-OCHA/hid-puppeteer — the "simple" flow is the only one available now, but it is sufficient to test. I'm still working on the "extra secure" flow and will ping again if I get it done before you review again.

orakili (approved):
Working well I think. When not logged in and I use one of those wrong URLs, I get redirected to the login page of the HID api site. When logged in, I see the new error message instead of the basic text one.

If the `client_id` and `redirect_uri` are correct I get redirected to the client site and for example see a "Login failed. Invalid OAuth2 state." error message on the client site depending on the other params.

rupl:
"I think" is as good as 💯  unless you're working with OAuth all the time 😂 

Thanks for the test!

PR #186, opened 2021-01-27
[OPS-6857] have docs generated automatically at build time; make git …

teodorescuserban:
…ignore docs artefact


rupl:
I didn't realize it would be so simple. Thanks!

PR #187, opened 2021-01-29
HID-2169: admin area to manage oauth clients

rupl:
# HID-2080

Sorry @orakili but one day I will stop throwing random fixes into bigger branches. But I had to include HID-2080 in order to finish my work for the admin UI.

`PUT /api/v3/client/{id}` should now produce `400 Bad Request` with validation errors in message body. Otherwise it is unchanged.

# HID-2169

@left23 the rest is admin UI stuff. If you can try all the usual things: log in as non-admin and try to access area, falsify forms, etc. Per our discussion (and largely Emma's wish) we intentionally went very light on form validation. However there are a few things the DB will complain about (duplicate client IDs, missing fields) and I think it handles all the show-stoppers.

@emmajane the bonus for you is the search, which I think works really nicely now. You can enter several words and it matches in any order. I made it search the client name, ID, urls, and description fields, so e.g. if you have someone's name in mind, that should turn up what you want:

Screen Shot 2021-01-29 at 10 38 30
Screen Shot 2021-01-29 at 10 39 00


emmajane:
Spiffy! Thanks, @rupl !!!

left23:
@rupl On first attempt as non-admin user, I get the expected error message when I try to visit /admin

left23 (approved):
@rupl Looks good.
I was able to edit OAuth client fields and was prompted by the required fields where relevant if I was missing data. I got a error message when I tried to use an existing Client ID.
I was able to change an existing Client to have the same name as another, but I don't really consider this an issue as it appears the crucial fields are all unique.
Approving.

rupl:
@left23 thanks! yes regarding the validation, I was asked not to "go overboard" as our intended audience is 3-4 people all in one flowdock channel. But all the validations you discovered are mission critical and/or required for DB integrity.

I think having the same name on two different clients might even be a "feature" (dev air quotes) since it allows us to facilitate transitions within active Partner sites that want to transition between two client IDs or secrets.

orakili (approved):
Reviewed changes related to HID-2080 and good to me.

PR #188, opened 2021-01-29
HID-2136: prevent 500 during access_token requests

rupl:
# HID-2136

Please check out the issue for a full description, but long story short I broke this part of HID for _some_ users (but not Drupal, which is why it was hard to track down). I'll comment inline with the major changes since I added tons of comments in the process of debugging.

## Testing

Since the "extra secure" flow is really difficult to mock, I have been using a local Drupal installation to test. I modified the port of HID to be 8080 and ran the drupal installation on port 80 without any modifications.

Once that is set up, log into Drupal as admin and adjust the `social_auth` config as desired to test various conditions (wrong client ID, wrong secret, etc).

### DIFF
--- a/api/controllers/AuthController.js
+++ b/api/controllers/AuthController.js
@@ -865,7 +865,7 @@ module.exports = {
             fail: true,
           },
         );
-        throw Boom.badRequest('Missing authorization code');
+        return Boom.unauthorized('Missing authorization code');
       }
 
       // Check client_id and client_secret
@@ -899,7 +899,7 @@ module.exports = {
             fail: true,
           },
         );
-        throw Boom.badRequest('invalid client authentication');
+        return Boom.unauthorized('invalid client authentication');
       }
 
       // Look up OAuth Client in DB.
@@ -916,7 +916,7 @@ module.exports = {
             },
           },
         );
-        throw Boom.badRequest('invalid client_id');
+        return Boom.badRequest('invalid client_id');
       }
 
       // Does the client_secret we received match the DB entry?
@@ -933,7 +933,7 @@ module.exports = {
             },
           },
         );
-        throw Boom.badRequest('invalid client_secret');
+        return Boom.unauthorized('invalid client_secret');
       }
 
       // Grab token and type.
@@ -955,10 +955,11 @@ module.exports = {
             },
           },
         );
+
         // OAuth2 standard error.
         const error = Boom.badRequest('invalid authorization code');
         error.output.payload.error = 'invalid_grant';
-        throw error;
+        return error;
       } else {
         logger.info(
           '[AuthController->accessTokenOauth2] Successful access token request',
@@ -985,7 +986,9 @@ module.exports = {
           stack_trace: err.stack,
         },
       );
-      return err;
+
+      // Send a documented error code back.
+      return Boom.badRequest(err.message);
     }
   },
 
--- a/api/controllers/AuthController.js
+++ b/api/controllers/AuthController.js
@@ -813,8 +813,10 @@ module.exports = {
         // @see HID-2156
         user.authorizedClients.push(request.yar.authorize[request.payload.transaction_id].client);
         user.markModified('authorizedClients');
+        await user.save();
+
         logger.info(
-          '[AuthController->authorizeOauth2] Added authorizedClient to user',
+          '[AuthController->authorizeOauth2] Added OAuth Client to user profile',
           {
             request,
             security: true,
@@ -827,7 +829,6 @@ module.exports = {
             },
           },
         );
-        await user.save();
       }
       const response = await oauth.decision(request, reply);
       return response;
--- a/api/controllers/AuthController.js
+++ b/api/controllers/AuthController.js
@@ -906,7 +906,7 @@ module.exports = {
       client = await Client.findOne({ id: clientId });
       if (!client) {
         logger.warn(
-          '[AuthController->accessTokenOAuth2] Unsuccessful access token request due to wrong client ID.',
+          '[AuthController->accessTokenOAuth2] Unsuccessful access token request due to non-existent client ID.',
           {
             request,
             security: true,
@@ -950,6 +950,7 @@ module.exports = {
             security: true,
             fail: true,
             oauth: {
+              client_id: clientId,
               code,
             },
           },
--- a/api/controllers/AuthController.js
+++ b/api/controllers/AuthController.js
@@ -846,10 +846,16 @@ module.exports = {
     }
   },
 
+  /**
+   * Issues access tokens during "Extra Secure" OAuth flows.
+   *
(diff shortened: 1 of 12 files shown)


api/controllers/AuthController.js
```diff
@@ -485,7 +511,30 @@ module.exports = {
...
+   *
+   * @see authorizeOauth2()
+   */
   async authorizeDialogOauth2(request, reply) {
+    // For some errors, we end up showing a prompt saying that the problem
+    // originated on the website which sent the user to HID. It's not always
+    // possible, but when we can we populate this URL so we can link them back
+    // to where they came from.
+    let errorRedirectUrl = '';
+
```
rupl, line 537: this (and any line which outputs this var) is actually part of HID-2117. Sorry.

api/controllers/AuthController.js
```diff
@@ -782,19 +865,21 @@ module.exports = {
             fail: true,
           },
         );
-        throw Boom.badRequest('Missing authorization code');
+        return Boom.unauthorized('Missing authorization code');
```
rupl, line 868: I am converting any error where we know the reason to `return Boom...` instead of throwing. The throws were causing us to return 500s in tons of situations where 400/401 were the appropriate response.

api/controllers/AuthController.js
```diff
@@ -856,23 +950,25 @@ module.exports = {
...
         logger.info(
           '[AuthController->accessTokenOauth2] Successful access token request',
           {
             request,
             security: true,
             oauth: {
-              client_id: request.payload.client_id,
-              client_secret: `${request.payload.client_secret.slice(0, 3)}...${request.payload.client_secret.slice(-3)}`,
+              client_id: clientId,
+              client_secret: `${clientSecret.slice(0, 3)}...${clientSecret.slice(-3)}`,
```
rupl, line 971: These two lines are the core bugfix. If you look above you see that `clientId` and `clientSecret` can be derived from two sources, per OAuth protocol, and we have docs describing both methods.

Since Drupal used the POST body method (which puts the data in `request.payload`) then all of our Drupal requests for `access_token` were passing through and getting logged just fine.

However if someone coded up a custom OAuth implmementation like in HID-2136, and they chose the Basic Auth header instead of POST body, the log itself was causing a 500 when in reality everything had been successful. My bad.

api/controllers/AuthController.js
rupl: In testing this on my local GHO, I didn't see any ill effects. There's still an ugly PHP error when config is wrong, but maybe you can provide some Drupal insight on how to prevent it. Here's a sample drupal error I made by falsifying the `client_secret` and attempting OAuth login:

```
The website encountered an unexpected error. Please try again later.

League\OAuth2\Client\Provider\Exception\IdentityProviderException: Bad Request in League\OAuth2\Client\Provider\HumanitarianId->checkResponse() (line 124 of /srv/www/vendor/un-ocha/oauth2-hid/src/Provider/HumanitarianId.php).
League\OAuth2\Client\Provider\AbstractProvider->getParsedResponse(Object) (Line: 537)
League\OAuth2\Client\Provider\AbstractProvider->getAccessToken(Object, Array) (Line: 31)
Drupal\social_auth_hid\HidAuthManager->authenticate() (Line: 250)
Drupal\social_auth\Controller\OAuth2ControllerBase->processCallback() (Line: 70)
Drupal\social_auth_hid\Controller\HidAuthController->callback()
call_user_func_array(Array, Array) (Line: 123)
Drupal\Core\EventSubscriber\EarlyRenderingControllerWrapperSubscriber->Drupal\Core\EventSubscriber\{closure}() (Line: 573)
Drupal\Core\Render\Renderer->executeInRenderContext(Object, Object) (Line: 124)
Drupal\Core\EventSubscriber\EarlyRenderingControllerWrapperSubscriber->wrapControllerExecutionInRenderContext(Array, Array) (Line: 97)
Drupal\Core\EventSubscriber\EarlyRenderingControllerWrapperSubscriber->Drupal\Core\EventSubscriber\{closure}() (Line: 151)
Symfony\Component\HttpKernel\HttpKernel->handleRaw(Object, 1) (Line: 68)
Symfony\Component\HttpKernel\HttpKernel->handle(Object, 1, 1) (Line: 57)
Drupal\Core\StackMiddleware\Session->handle(Object, 1, 1) (Line: 47)
Drupal\Core\StackMiddleware\KernelPreHandle->handle(Object, 1, 1) (Line: 106)
Drupal\page_cache\StackMiddleware\PageCache->pass(Object, 1, 1) (Line: 85)
Drupal\page_cache\StackMiddleware\PageCache->handle(Object, 1, 1) (Line: 47)
Drupal\Core\StackMiddleware\ReverseProxyMiddleware->handle(Object, 1, 1) (Line: 52)
Drupal\Core\StackMiddleware\NegotiationMiddleware->handle(Object, 1, 1) (Line: 23)
Stack\StackedHttpKernel->handle(Object, 1, 1) (Line: 708)
Drupal\Core\DrupalKernel->handle(Object) (Line: 19)
```

orakili: @rupl The `League\OAuth2` library throws exceptions when something is wrong but the drupal `social_api`, `social_auth` and `social_auth_hid` modules don't seem to try to catch them.

After a little debugging, the error sent by the HID API looks like: 400 Bad Request with the a json payload containing the `message`: "invalid client_secret". 

However according to [some docs](https://www.oauth.com/oauth2-servers/server-side-apps/possible-errors/), it looks like the payload should have a `error_description` property instead of (or in addition to) the `message` one. 

This is also what the `un-ocha/oauth2-hid` is expecting:

```php
    protected function checkResponse(ResponseInterface $response, $data)
    {
        if (isset($data['error'])) {
            throw new IdentityProviderException(
                $data['error_description'] ?: $response->getReasonPhrase(),
                $response->getStatusCode(),
                $response
            );
        }
    }
  ```
  
 I would propose to change that if possible. That would not change the fact that the `social_*` modules don't handle the exceptions but at least might be closer to the specs.
 
 As for preventing the PHP error, I think we could add some exception catching to the `social_auth_hid` module and redirect to the homepage or something like that and log the error message? cc @attiks

orakili (approved):
Looks good to me.

api/controllers/AuthController.js
rupl: oh hey that's great feedback! In the `hapi-oauth2orize` plugin, there's a function `transformBoomError()` which appends that `error_description` field automatically. It sounds like I should be processing all our errors in this function through that transformer, in order to produce OAuth-compatible errors. I'll give it a shot.

rupl: (just writing some notes to myself since this is getting de-prioritized)

In addition to the error format, the [OAuth possible errors](https://www.oauth.com/oauth2-servers/server-side-apps/possible-errors/) doc from Jonathan mentions some redirects when they are both possible and safe. Can't hurt to implement those too.

PR #189, opened 2021-02-02
HID-2152: remove outdated/legacy domains from allow-list

rupl:
# HID-2152

Just one step of the process. Not urgent, does not need to be in sync with a particular deploy.

### DIFF
--- a/api/services/HelperService.js
+++ b/api/services/HelperService.js
@@ -8,36 +8,28 @@ const queryOptions = [
   'sort',
   'fields',
 ];
+
+// TODO: move to env var
+//
+// @see https://humanitarian.atlassian.net/browse/HID-1894
 const authorizedDomains = [
   // Production
-  'https://humanitarian.id',
   'https://api.humanitarian.id',
   'https://auth.humanitarian.id',
 
-  // Dev
-  'https://dev.humanitarian-id.ahconu.org',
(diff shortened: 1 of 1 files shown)


api/services/HelperService.js
```diff
@@ -8,36 +8,28 @@ const queryOptions = [
...
+
+  // Dev
+  'https://dev.api-humanitarian-id.ahconu.org',
+  'https://dev.auth-humanitarian-id.ahconu.org',
 
   // Local
-  'https://app.hid.vm',
   'https://api.hid.vm',
-  'http://app.hid.vm',
+  'https://auth.hid.vm',
```
teodorescuserban, line 30:
```suggestion
  'http://auth.hid.vm',
```

teodorescuserban (approved):
with http on local i think it looks good.

api/services/HelperService.js
rupl: I have all four listed already...

api/services/HelperService.js
```diff
@@ -8,36 +8,28 @@ const queryOptions = [
...
   'http://api.hid.vm',
-  'http://v3.hid.vm',
-  'http://admin.hid.vm',
-  'http://hid.vm',
+  'http://auth.hid.vm',
```
rupl, line 32: @teodorescuserban

PR #190, opened 2021-02-15
OPS-7221: upgrade to node.js 14

rupl:
# OPS-7112

I think this gets us on node.js 14 doing as little as possible to get the site running.

@teodorescuserban I dipped into the container and manually installed python3 to get past an error initially, but then I might have deleted the library calling for it, the `sharp` dependency.

Would it be possible for you to do a full build of this to see if it actually works, rather than only looking at the code?


teodorescuserban (approved):
lgtm

PR #192, opened 2021-02-18
HID-2190: create new OAuth client via jenkins command

rupl:
# HID-2190

we want to create new OAuth clients via jenkins. this new script at `commands/createOAuthClient.js` does that.

Testing locally is done by spinning up the container then running:

```
docker-compose exec dev node commands/createOAuthClient.js
```

It should log success or failure. Searching the list for "change" at http://api.hid.vm/admin will turn up the new entry like so:

Screen Shot 2021-02-18 at 15 28 12


orakili (approved):
It works well. I  created a few clients and tested on different local drupal sites after changing their settings to use the new clients.

PR #194, opened 2021-02-19
HID-1972: remove Notifications from HID

rupl:
# HID-1972

The PR completely removes the concept of Notifications from HID. The one file that remains is intentional: the Notification.js which defines the DB schema. That file is needed in order to run the migration which drops the collection (AKA table)

The intention is to run `commands/dropNotification.js` from Jenkins once this branch makes it to production.

I also scrubbed any mention of v2 from the README. It will continue getting smaller on a branch-by-branch basis as I work through these tickets.

## Testing

```
$ docker-compose exec dev node commands/dropNotification.js
info: [commands->dropNotification] dropped Notification collection level=info, @timestamp=2021-02-19T13:55:40.055Z, ,

> show tables;
// it should not list the `notification` table anymore
> db.notification.count()
0
```

Aside from running the migration as demonstrated above, doing an OAuth login from a Drupal site should be sufficient, as that is the only remaining aspect of HID which is supported.


orakili (approved):
Works as advertised. The `dropNotifications.js` did remove the table and I could still log in my local sites.

PR #195, opened 2021-02-19
HID-1938: remove authOnly field

rupl:
# HID-1938

Removes one field `authOnly` from all user records. This PR adds a command that will be run via Jenkins.

## Testing

```
$ docker-compose exec dev node commands/removeFieldAuthOnly.js
info: [commands->removeFieldAuthOnly] Removed authOnly field from all user objects level=info, @timestamp=2021-02-19T15:05:19.340Z, ,

> db.user.count({authOnly: true})
0
> db.user.count({authOnly: false})
0
```

Obviously, logging into a Drupal site should still work.

Furthermore, editing a user profile, or logging in with HID, or managing any of your settings shouldn't restore the field to the user. I had a little trouble with this one and do want to confirm it, but we'll also have dev to check.


orakili (commented):
The `removeFieldAuthOnly.js` did what it was intended to do. I could still log in local sites and updating my profile (name change) and settings (password change) didn't cause the field to be added back the field.

orakili (approved):
Looks like I select the wrong option. Approved.

PR #197, opened 2021-02-22
HID-1932: appMetadata migration

rupl:
# HID-1932

We're "migrating" this field to `dev/null` wholesale. It was used by the AngularJS codebase to store info that should have been cookied in the browser.

```
$ docker-compose exec dev node commands/removeFieldAppMetadata.js
info: [commands->removeFieldAppMetadata] Removed appMetadata field from all users level=info, @timestamp=2021-02-22T08:52:09.180Z, ,

> db.user.count({appMetadata: {$exists: true}})
0
```


orakili (approved):
Works as advertised. Could still log in and the command did remove the field.

PR #199, opened 2021-02-22
HID-1914: remove status field

rupl:
# HID-1914

Removes `status` field. It's a little harder to grep the codebase for the word "status" than some of these other fields, but I feel like I removed all the instances within branches of code that we're ultimately keeping.

There are some leftover references in functions that we know will disappear, such as CSV export. But those cases are still safe because the field was always optional and it defaults to empty string when the value is falsey.

```
$ docker-compose exec dev node commands/removeFieldStatus.js
info: [commands->removeFieldStatus] Removed status field from all users level=info, @timestamp=2021-02-22T13:42:59.003Z, ,

> db.user.count({status: {$exists: true}})
0
```

As usual: OAuth needs to function like normal. If that's still working then there's not much else to check.


orakili (approved):
Working well. Could still log in after removing this field.

PR #200, opened 2021-02-22
HID-1920: remove voips field from user profile

rupl:
# HID-1920

Removes the `voips` field with the same caveat as #199 that several instances remain in the codebase, but will be eliminated on their own at a later point. In this case the `voips` field doesn't seem properly guarded against calls to it (i.e. `if (user.voips.length)` would cause errors in JavaScript, but the CSV export is no longer possible to invoke since HID Contacts (the AngularJS app) is now offline.

```
$ docker-compose exec dev node commands/removeFieldVoips.js
info: [commands->removeFieldVoips] Removed voips field from all users level=info, @timestamp=2021-02-22T14:25:50.683Z, ,

> db.user.count({voips: {$exists: true}})
0
```

OAuth should still work as usual.


orakili (approved):
Working as expected.

PR #201, opened 2021-02-22
HID-1916 + HID-1917: remove fields related to phone numbers

rupl:
# HID-1916 + HID-1917

Finally removing a dependency in addition to some fields! This PR drops `phone_number`, `phone_number_type`, `phone_number_verified`, and `phonesVisibility` from the user profile. Same caveat as #199 and #200 that CSV export would probably be crashy, but we don't care for now.

```
$ docker-compose exec dev node commands/removeFieldPhoneNumbers.js
info: [commands->removeFieldPhoneNumbers] Removed fields related to phone numbers from all users level=info, @timestamp=2021-02-22T14:53:06.093Z, ,

> db.user.count({phonesVisibility: {$exists: true}})
0
> db.user.count({phone_numbers: {$exists: true}})
0
> db.user.count({phone_number_type: {$exists: true}})
0
> db.user.count({phone_number_verified: {$exists: true}})
0
```

OAuth should still work like usual.


orakili (approved):
Working good as well.

PR #203, opened 2021-02-23
HID-1919: remove emailsVisibility field from all users

rupl:
# HID-1919

Removes the `emailsVisibility` field. Field remains in CSV export and GSSSync functions.

```
$ docker-compose exec dev node commands/removeFieldEmailsVisibility.js
info: [commands->removeFieldEmailsVisibility] Removed emailsVisibility field from all users level=info, @timestamp=2021-02-23T10:09:34.514Z, ,

> db.user.count({emailsVisibility: {$exists: true}})
0
```

OAuth should still work like usual.

### DIFF
--- a/api/models/User.js
+++ b/api/models/User.js
@@ -265,11 +265,6 @@ const UserSchema = new Schema({
     type: [emailSchema],
     // readonly: true
   },
-  emailsVisibility: {
-    type: String,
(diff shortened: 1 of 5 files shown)


orakili (approved):
Command works well and I could still log in.

PR #205, opened 2021-02-23
HID-1939: remove location, locations, locationsVisibility fields

rupl:
# HID-1939

Removes the `location`, `locations`, and `locationsVisibility` fields from all user profiles. References remain in Outlook, GSS, and CSV export functions.

```
$ docker-compose exec dev node commands/removeFieldLocations.js
info: [commands->removeFieldLocations] Removed location, locations, locationsVisibility fields from all users level=info, @timestamp=2021-02-23T10:43:01.206Z, ,

> db.user.count({location: {$exists: true}})
0
> db.user.count({locations: {$exists: true}})
0
> db.user.count({locationsVisibility: {$exists: true}})
0
```

OAuth should continue working as usual.

### DIFF
--- a/commands/sendReminderCheckinEmails.js
+++ b/commands/sendReminderCheckinEmails.js
@@ -29,7 +29,7 @@ async function run() {
         && offset < 72 * 3600 * 1000
         && !lu.deleted) {
         const hasLocalPhoneNumber = false;
-        const inCountry = await user.isInCountry(lu.list.metadata.country.pcode);
+        const inCountry = false;
         lu.remindedCheckin = true;
         await user.save();
       }
--- a/commands/sendReminderCheckinEmails.js
+++ b/commands/sendReminderCheckinEmails.js
@@ -28,7 +28,6 @@ async function run() {
       if (!lu.remindedCheckin && offset > 48 * 3600 * 1000
         && offset < 72 * 3600 * 1000
         && !lu.deleted) {
-        const hasLocalPhoneNumber = false;
         lu.remindedCheckin = true;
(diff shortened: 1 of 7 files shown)


commands/sendReminderCheckinEmails.js
```diff
@@ -29,7 +29,7 @@ async function run() {
         && offset < 72 * 3600 * 1000
         && !lu.deleted) {
         const hasLocalPhoneNumber = false;
-        const inCountry = await user.isInCountry(lu.list.metadata.country.pcode);
+        const inCountry = false;
```
orakili, line 32: I think that could be removed as well as the other references of `inCountry` in https://github.com/UN-OCHA/hid_api/blob/e29f471b98547758573d0e9fb244fad34d9f9101/emails/reminder_checkin/html.ejs#L9 and https://github.com/UN-OCHA/hid_api/blob/e29f471b98547758573d0e9fb244fad34d9f9101/emails/reminder_checkin/fr/html.ejs#L9

orakili (approved):
Works well and good to get rid of the use on the HR.info API. There are a bit of leftovers for `inCountry` in the reminder templates that I think could be removed as well.

commands/sendReminderCheckinEmails.js
rupl: Whoops, perhaps I could have set the var to `true` to ensure the email never displays that part of the message.

At any rate, we're dropping all of the check-in code "soon" (I might get to it today) so by the time we deploy these changes, it will get expunged.

PR #206, opened 2021-02-23
HID-2193: remove functional_roles field from all users

rupl:
# HID-2193

Removes the `functional_roles` field from User schema, and code remains in the CSV, GSS, Outlook vestigial functions.

```
$ docker-compose exec dev node commands/removeFieldFunctionalRoles.js
info: [commands->removeFieldFunctionalRoles] Removed functional_roles field from all users level=info, @timestamp=2021-02-23T12:50:55.829Z, ,

> db.user.count({functional_roles: {$exists: true}})
0
```

OAuth should still work.

Feeling like I should tackle these GSS and CSV things so I can stop sounding like a broken record 😆

### DIFF
--- a/api/models/User.js
+++ b/api/models/User.js
@@ -386,7 +386,6 @@ const UserSchema = new Schema({
       message: 'HTML in job titles is not allowed',
     },
   },
-  functional_roles: [listUserSchema],
   // TODO: figure out validation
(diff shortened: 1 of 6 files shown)


orakili (approved):
:+1:

PR #208, opened 2021-02-23
HID-1933: remove connections field from all users

rupl:
# HID-1933

Remove `connections` field from all users. There's also some emails being deleted in addition to API methods to manage connections since this was previously a social network type feature of HID Contacts.

```
$ docker-compose exec dev node commands/removeFieldConnections.js
info: [commands->removeFieldConnections] Removed connections field from all users level=info, @timestamp=2021-02-23T14:41:13.631Z, ,

> db.user.count({connections: {$exists: true}})
0
```

The following should be tested:

- OAuth logins
- HID logins
- Profile edits
- Deleting an HID account — I recommend using `db.user.findOne()` on the mongo console, resetting the password of someone else, then deleting their account. However if you delete your own there's nothing stopping you from signing back up using the same address.


orakili (approved):
Tested login into API, editing profile, editing password, deleting account, creating account and login into drupal site. No problem.

PR #210, opened 2021-02-25
Remove legacy API surface

rupl:
I probably missed some tickets, but all of these are covered in one form or another:

- HID-1934
- HID-1946
- HID-1947
- HID-1948
- HID-1950
- HID-1953
- HID-1979
- HID-2007
- HID-2193

Lots of files totally removed, namely Controllers and Services. None of the Models were removed yet, and that's necessary in order to run the migrations where the model itself gets dropped. I'll do a cleanup PR once we go to production and the data is gone. If a Model does show deletions, it's most likely that I did it to completely drop a dependency, which is pleasantly down to one low-level vulnerability according to `npm audit` 🎉 

In terms of the wide scope here, you have to believe me that it was more or less necessary to do about two-thirds of it at the same time. The rest of it was just ripping the band-aid off, so to speak. I removed some files (like email templates) that were already vestigial, and for instance there's no `dropOutlook` because that seems to have already happened in the past. The code that remained was just cruft.

Here's the list of commands that need to be run using the same convention as previous PRs:

```
# Dropping whole tables
commands/dropGSSSync.js
commands/dropList.js
commands/dropOperation.js
commands/dropService.js
commands/dropServiceCredentials.js

# Removing fields from user profiles
commands/removeFieldGssOutlook.js
  - googleCredentials
  - outlookCredentials
commands/removeFieldLists.js
  - organization
  - organizations
  - lists
  - operations
  - bundles
  - disasters
  - offices
commands/removeFieldSubscriptions.js
  - subscriptions
```

## Testing

Like always, our core functionality is a user logging in, editing profile/emails/settings, and using HID to log into Drupal.

However, the entire remaining functionality could use a good functional test. I found so many edge cases while testing so anything you can imagine doing is appreciated as a test (e.g. deleting an email used to trigger GSSSync, because of course). 

I kept my eye on how the v3 API functions too, so I'm feeling good about it, but again any wild ideas you have to run bizarre API calls or try and break it will also be appreciated.


rupl:
@left23 if you have the opportunity on monday, I'd appreciate if you could functionally review this work and confirm that most features are still working on your local. I'd be happy to pair and help you run migrations and stuff.

left23 (approved):
Functional testing is all good by me.
I was able to login with an existing account and change password and edit profile info. I was able to add, confirm and remove secondary emails, and set an email as primary address. I can remove authorised apps. I can create a new account, enable and disable 2FA, reset password, also with 2FA prompts, and delete my account. 
As far as I can tell, the functionality remains as expected.
Approving. Nice work @rupl

rupl:
Thanks so much @left23 — @orakili the only remaining test is Drupal Oauth logins, plus code review for files which aren't being completely deleted. I hope this makes the review process a bit less time-consuming.

orakili (commented):
@rupl, I reviewed the code and tested Drupal logins. However I had issues when running the `drop*` commands, probably due to an outdated DB that suffered from last weeks PRs. Once I can get my hands on a DB snapshot, I'll try again. If you need to merge that to move forward, please go ahead. It looks good to me.

rupl:
What were the errors, just out of curiosity?

rupl:
That seems to be a [DB connection error](https://stackoverflow.com/questions/37136204/mongoerror-ns-not-found-when-try-to-drop-collection). Are all of the scripts to drop tables producing the same error?

orakili (approved):
Tested again with fresh dump of the DB and all the commands are working and was still able to log in drupal sites, edit profile, password etc.

Not related to this PR and I didn't pay attention before but I find the UI a bit confusing with the `<button>` elements used for tags like `primary` with a change of background color on hover. They feel like something should happen when clicking on them. Also semantically that doesn't sound correct.

Related to that, the cursor for the buttons is not consistent. For example for the "login" button on the homepage, the cursor is not changed to "pointer" while it is for the "edit profile" button. Using the pointer cursor for all the buttons that do something would probably help distinguish them from the tags.

Screen Shot 2021-03-04 at 14 03 50
Screen Shot 2021-03-04 at 13 37 26

Let me know if you want a ticket for that.

rupl:
All good points and this mis-use of buttons was precisely why I chose to convert the GHO appeal tags as my CD component assignment in January. Neither I nor @left23 liked the idea of using a button for non-interactive labels, but we used what was on hand since we needed this UI, and HID v3 was supposed to launch rather rapidly. Afterwards I wanted to get the Appeal tags into CD so that we could use them as labels in HID. What you're seeing is indeed unfinished business.

The message about `[confirm]` is easy to fix. I have been meaning to do it and since it's a small templating change, I'll ticket and add it to my current backlog because I agree it's confusing. HID-2201

PR #211, opened 2021-02-26
OPS-7225: update domain list to complement recent stack updates

rupl:
# OPS-7225

minor housekeeping to keep up with latest stack updates.


rupl:
@left23 I'm assigning this to you for review since you use the stack. with this branch it should be safe to pull on the stack's `main` branch and check this PR out, then everything should continue working as usual.

Things to test: password resets, email confirmations

left23:
I pulled `main` from `hid_stack` and checked out this branch.
Based on the PR diff, I'm not sure if I should go to `https://hid.test/` or the usual `http://api.hid.vm` - the latter still works, the former does not (Server not found)

On `http://api.hid.vm` I was able to login fine with my regular email address, but when I reset the password, I get
`There was an error resetting your password.`
and `warn: [UserController->resetPasswordEndpoint] app_reset_url http://api.hid.vm/new_password is not in authorizedDomains allowlist ...`
I also cannot register `warn: [UserController->resetPasswordEndpoint] app_reset_url http://api.hid.vm/new_password is not in authorizedDomains allowlist ...`

left23 (approved):
Second attempt after adding `hid.test` to hosts file
I can reset my password and I get an email confirmation, and I can log in.
I also get email confirmation when registering.
Approving.

PR #212, opened 2021-02-26
HID-2107: change verb for password reset

rupl:
# HID-2107

`PUT` is expected to have identical outcomes when the same API call is made more than once. Since the password reset will fail the second time due to the current password being different, `PUT` is inappropriate.


orakili (approved):
👍

PR #209, opened 2021-02-23
HID-1926: remove fields related to verified status

rupl:
# HID-1926

Removes the following fields:

- `verified`
- `verified_by`
- `verifiedOn`
- `verificationExpiryEmails`

And the following Jenkins command: `unverifyAfterOneYear`

```
$ docker-compose exec dev node commands/removeFieldVerified.js
info: [commands->removeFieldVerified] Removed verified, verified_by, verifiedOn, and verificationExpiryEmail fields from all users level=info, @timestamp=2021-02-23T15:24:36.852Z, ,

$ docker-compose exec db mongo local
> db.user.count({verified: {$exists: true}})
0
> db.user.count({verified_by: {$exists: true}})
0
> db.user.count({verifiedOn: {$exists: true}})
0
> db.user.count({verificationExpiryEmail: {$exists: true}})
0
```

The following should be tested:

- OAuth login via Drupal site
- New account registration
- Profile edits

PR #213, opened 2021-02-26
HID-1927: drop trusteddomain collection

rupl:
# HID-1927

I didn't scrub thoroughly enough in #209 so here's a second pass. Drops the `TrustedDomain` controller and associated DB table. There's also a command being dropped from Jenkins.


orakili (approved):
Command worked well and I could still log into drupal sites.

PR #214, opened 2021-02-26
HID-2196: drop the Duplicate and Result collections

rupl:
# HID-2196

Drops the `duplicate` and `results` tables from the DB. I had to actually create the node.js models because these tables were from HID v1 if I read little odds and ends in both codebase and Jira. So I made a blank schema in order to connect to the table and drop it.

```
$ docker-compose exec dev node commands/dropDuplicate.js
info: [commands->dropDuplicate] dropped Duplicate collection level=info, @timestamp=2021-02-26T12:58:52.994Z, ,

$ docker-compose exec dev node commands/dropResult.js
info: [commands->dropResult] dropped Result collection level=info, @timestamp=2021-02-26T12:59:13.044Z, ,

> show tables;
client
flood
jwttoken
oauthtoken
startup_log
user
```

There's basically nothing to test, because these tables weren't even accessible from node.js before I made this branch with the two new model files.

This concludes the DB pruning! (I think!!)


orakili (approved):
Commands are working well and I could still log into drupal sites.

PR #215, opened 2021-03-02
HID-2025: upgrade hapi@20

rupl:
# HID-2025

Upgrading hapi to the latest community-supported version: v20

There was one breaking change related to logging IPs on aborted requests (which happens when an auth token is missing or invalid) — other than that it was smooth sailing ⛵ 

I fixed the mongo deprecation warning, removed a little bit of v2 code, and added a script to make sure our API docs stay current without having to manually update.

## Testing

PR requires `npm install` to upgrade all the packages, but then testing Drupal logins should be enough. We are testing the interface on other branches frequently enough.

Running `npm run docs` will update the docs and the correct version should appear in the Swagger page.


config/web.js
```diff
@@ -195,6 +195,12 @@ module.exports = {
   },
 
   options: {
+    info: {
+      // We always want access to request.info.remoteAddress in hapi@19+
+      //
+      // @see https://github.com/hapijs/hapi/issues/4017
+      remote: true,
+    },
```
rupl, line 203: This fixed the breaking change and retained the pre-v19 behavior of always populating `request.info.remoteAddress` even after the request is aborted.

docs/swaggerBase.yaml
```diff
@@ -1,7 +1,7 @@
 openapi: 3.0.3
 x-api-id: hid-api
 info:
-  version: 3.0.5
+  version: <<<VERSION>>>
```
rupl, line 4: this got way out of date in the last couple weeks. Since we were already generating the docs at deploy-time, I figured why not keep the version in sync automatically too?

orakili (approved):
Working well, no problem running the update and could still log into drupal sites. No problem either with the command to recreate the documentation and it shows the proper version.

PR #217, opened 2021-03-03
HID-2066: avoid 500s in email validation function

rupl:
# HID-2066

The function needed pristine input or it would 500 in a bunch of cases. It seems to be fixed up, with good API calls still sending off the email, while bad requests (missing payload entirely, missing `app_validation_url`, wrong url) all get 400s now. It used to 500 on the missing payload/url.

Here's a valid cURL to run. Test by removing `--data` or sending invalid JSON, removing the URL, sending a bad URL, etc. 

```
curl --request PUT \
  --url http://hid.test:8080/api/v3/user/emails/ruppel@reliefweb.int \
  --data '{
  "app_validation_url": "http://hid.test:8080/verify2"
}'
```

Code-wise, I changed `UserController.validateEmail()` but you can ignore the first half of the function which is only used once the email is sent. Look for the comment that begins with "When hash wasn't present" to find the code relevant to the PR.


orakili (approved):
Works well. Tested different scenarios and got the `{"statusCode":400,"error":"Bad Request","message":"Invalid app_validation_url"}` message.

Also tested email verification through the UI and it was still working.

PR #218, opened 2021-03-03
HID-2068: avoid 500s on resetPasswordEndpoint

rupl:
# HID-2068

Avoiding more 500s on API methods. Here's another cURL to hopefully make things simple. Again like #217 altering this cURL in any way should result in 400 instead of 500. Seeing 500 means I have work to do, and the API server error will help me know what to fix.

```
curl --request PUT \
  --url http://hid.test:8080/api/v3/user/password \
  --data '{
	"email": "ruppel@reliefweb.int",
	"app_reset_url": "http://hid.test:8080/new_password"
}'
```


orakili (approved):
Working well. Always getting `{"statusCode":400,"error":"Bad Request","message":"app_reset_url is invalid"}` now whether the url is missing, invalid etc.

I also could still log in drupal sites.

PR #220, opened 2021-03-03
HID-2184: less definitive messaging around password resets

rupl:
# HID-2184

We want to make it clear the form submitted ok, but our internal business logic is dependent on a couple conditions, one of which is a security hole to disclose as user feedback within the message.

Therefore, we need to make it clear that the message _probably_ gets sent to your address, instead of implying it already got sent.

## Testing

Start logged out. Click on password reset and enter a fake email. You should see the message reflected in the diff. Now reset the password on a real email address. Same message should be shown.


left23 (approved):
I confirm the same message shows in both scenarios. Approving.

PR #221, opened 2021-03-03
HID-2185: password reset for validated secondary email addresses

rupl:
# HID-2185

We want to allow passwords to be reset from secondary emails which already answered the validation challenge.

## Testing

- ✅ Password reset for **unvalidated primary** address **should still work**
- ✅ Password reset for **validated primary** address **should still work**
- ✅ Password reset for a **validated secondary** address **should work**
- ❌ Password reset for an **unvalidated secondary** address **should NOT work**

In all cases, the "success" confirmation should be displayed. See #220 where we change the wording of the message to make it less definitive. If an email doesn't get sent, the API logs internally about why things failed. We don't want to be disclosing that to the end user on grounds of security.

I had to thoroughly check on case number 1 with unvalidated primary emails, but it is indeed how HID already operates and IMO makes sense. Otherwise if your spam swallowed the initial confirmation email, you would have no way of confirming your primary email address.


left23 (approved):
I can confirm the password reset doesn't work for unvalidated secondary emails, as is intended.

I was able to reset my password using the validated secondary email address.

I can confirm password reset for primary email address remains functioning. Approving.

PR #222, opened 2021-03-04
HID-2201: only show message about confirm buttons when they are present

rupl:
# HID-2201

We have some instructions about a <kbd>CONFIRM</kbd> button on the profile edit page. The instructions were confusing when the button they referred to was not present. Now they only show when needed.

### No Message

HID-2201-no-msg

### Message

HID-2201-message


left23 (approved):
Approving. Nice improvement for clarity.

PR #223, opened 2021-03-09
HID-2205: better log message for password reset

rupl:
# HID-2205

When someone attempts password reset with non-existent email, this will help us scan logs more easily


emmajane (approved):
Tasty log messages are tasty. Thanks @rupl !!

PR #224, opened 2021-03-10
HID-2118: Add org/env to OAuth Client config

rupl:
# HID-2118

We currently have a naming convention on most of our properties that follows: `org - Site name - env` — for a couple reasons we are changing the org/env to be separate fields:

- The system was mostly working but it's getting a bit confusing
- Certain end users were curious what "prod" means and we felt it best to start hiding that info from folks as it sounds strange when you aren't dealing with dev/test/prod all day.

The new fields are available in this PR to be used on an optional basis, allowing a gradual transition as we do data entry. 

## Testing

- Log in as an HID Admin and see if the new fields work as expected on the OAuth Client editing form. You will go to "Admin" in the top dropdown and edit any client, but keep in mind you have to test the Authorize form so try at least one that you have configured locally, e.g. GHO Local.
- Log in as an HID user and use the CD Header dropdown to visit **Settings** where you will see a list of clients. The "x" button now says "Revoke" and it should work like normal. Additionally, after you've edited a few Clients as Admin, those same clients should list any combination of metadata that you entered with one exception: environments of "Production" or "Prod" are not shown to the user.
- Log in via a Drupal site and see if the form mentions that the environment is for development/testing. We do a simple string check and look for "Prod" and failing to find it means we consider it for development.

Some screenshots that hopefully make it clearer:

HID-2118-authorize
HID-2118-user-settings


api/models/Client.js
```diff
@@ -84,6 +84,18 @@ const ClientSchema = new Schema({
       message: 'HTML code is not allowed in description',
     },
   },
+  organization: {
+    type: String,
+    required: false,
+    default: '',
+    maxlength: 16,
```
orakili, line 91: No luck for "Humanitarian Response" and the like ;-)

> Error [ValidationError]: Client validation failed: organization: Path `organization` (`Humanitarian Response`) is longer than the maximum allowed length (16).

templates/authorize.html
```diff
@@ -11,6 +11,16 @@ <h1 class="page-header__heading">Authorize <%= client.name %>?</h1>
         <li>Your email address<%= user.emails && user.emails.length > 1 ? 'es' : '' %></li>
         <li>Your HID user ID</li>
       </ul>
+      <p>
+        <em>
+          <% if (!!client.organization) { %>
+            <strong><%= client.name %></strong> is managed by <strong><%= client.organization %></strong>.<br>
+          <% } %>
+          <% if (!!client.environment && client.environment.indexOf('Prod') === -1) { %>
```
orakili, line 19:
```suggestion
          <% if (!!client.environment && client.environment.toLowerCase().indexOf('prod') === -1) { %>
```

templates/settings.html
```diff
@@ -34,22 +35,32 @@ <h2>Authorized Apps</h2>
...
-                    <svg class="cd-icon cd-icon--cancel" width="16" height="16" aria-hidden="true" focusable="false">
-                      <use xlink:href="#cd-icon--cancel"></use>
-                    </svg>
-                  </button>
+                  >Revoke</button>
+                </span>
+                <span class="client__info">
+                  <span class="client__name"><%= client.name %></span>
+                  <span class="client__meta">
+                    <% if (!!client.environment && client.environment.indexOf('Prod') === -1) { %>
```
orakili, line 52:
```suggestion
                    <% if (!!client.environment && client.environment.toLowerCase().indexOf('prod') === -1) { %>
```

orakili (approved):
Working good, but I would make the test on `Prod` case insensitive (see screenshot below) and also maybe increase the limit for the organization field to at least 32 characters to be a bit more flexible. (See individual comments).

Screen Shot 2021-03-12 at 13 42 49

api/models/Client.js
rupl: Whoops, I guess in that case I should have added the validation to the `<input>` as well. But in general point taken and we can bump to 32.

Per standup discussion we'll convert the Environment to be a `<select>` while keeping DB as a string of open values. I figure if we want to add new options it will be uncomplicated to modify HTML and nothing else.

PR #225, opened 2021-03-12
HID-2024: winston@3 upgrade

rupl:
# HID-2024

Winston 3 had some breaking changes and as a result we had to [refactor the log rewriter into the new thing called a formatter](https://github.com/winstonjs/winston/blob/master/UPGRADE-3.0.md#migrating-filters-and-rewriters-to-formats-in-winston3). What came with the breaking change is the inconvenience that formatters continually mutate the same object as it gets passed along, which was incompatible with our old formatter, which was sanitizing important data before writing the log (e.g. client secrets, specific payload fields like passwords, etc)

Most of the changes are find/replace inside the formatter itself, since the mutations were required to happen on the `info` that gets passed in, but we didn't want those changes affecting the `request` object when it gets used again after being logged. 

## Testing

You know the drill, maintaining drupal logins are top priority. Specifically, the logs should still show obfuscated client secrets such as `ABC...XYZ`, but continue functioning which indicates that the original `request` containing the unmodified secret got passed along in the process.

I have already tested other things like password resets via email and noted that all our sanitization seems to be happening to unset `password` and other tokens in payloads. However a second check never hurts so any form which submits sensitive info is a good thing to check:

- `/register`
- `/new_password`
- `/settings/password`

And the API version (which we recently modified):

```
curl --request POST \
  --url http://hid.test:8080/api/v3/user/YOUR_UID/password \
  --header 'Authorization: Bearer YOUR_TOKEN' \
  --data '{
	"new_password": "NEW_PASS",
	"old_password": "OLD_PASS"
}'
```

In this case if you see in the logs there should be a `Bearer` token which normally has three sections, but the log should only contain the first two (which can be base64 decoded for debugging purposes after extracting from Kibana) but the LAST section should be scrubbed, meaning the signature which validates the first two payloads is not logged.

### DIFF
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -1096,7 +1096,7 @@ module.exports = {
       {
         request,
         user: {
-          userId,
+          id: userId,
           email: record.email,
         },
       },
--- a/config/env/development.js
+++ b/config/env/development.js
@@ -1,8 +1,9 @@
 const winston = require('winston');
 const { hidFormatter } = require('../logs');
-require('winston-daily-rotate-file');
+const DailyRotateFile = require('winston-daily-rotate-file');
 
 module.exports = {
+  env: 'development',
   database: {
     stores: {
       development: {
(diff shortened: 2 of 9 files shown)


rupl:
@orakili you can skip this one, Lisa and I reviewed it together.

left23 (approved):
I can confirm the logs show obfuscated client secrets and bearer token, and I cannot see any other sensitive data in the logs.
Approving.

PR #227, opened 2021-03-15
HID-2173: add some about text to login page

rupl:
# HID-2173

Adding some "about this site" text to the login form, which currently functions as homepage. People who log into other sites with HID will also see this text next to or under the login form.

@left23 I had a question about the Grid in flowdock but though I'd tag you as a reviewer so you can advise me on specific lines if I am implementing the grid wrong. I'll leave a comment on my workaround I added.

### DIFF
--- a/assets/css/login.css
+++ b/assets/css/login.css
@@ -1,0 +1,9 @@
+/**
+ * Login page
+ */
+.page--login {}
+
+.page--login .cd-grid > * {
+  display: block;
+  background: transparent;
+}
--- a/templates/login.html
+++ b/templates/login.html
@@ -1,9 +1,11 @@
 <% include header %>
+<style><% include ../assets/css/cd-grid.css %></style>
+<style><% include ../assets/css/login.css %></style>
 
-<div class="api-page">
-  <main role="main" id="main-content" class="cd-container">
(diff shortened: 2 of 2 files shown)


assets/css/login.css
```diff
@@ -0,0 +1,9 @@
+/**
+ * Login page
+ */
+.page--login {}
+
+.page--login .cd-grid > * {
+  display: block;
+  background: transparent;
+}
```
rupl, line 9: When I created the grid, I noticed that the inner contents of my two grid columns took on a flexbox layout which was quite disruptive until I unset it. I did review the [CD Grid demo](https://feature.commondesign-unocha-org.ahconu.org/demo#cd-grid) and the styles are the same there as this branch, with the grid items using a flex layout with `row nowrap` and having a gray background.

Is my workaround the simplest fix at the moment?

left23: Ya, if it works, stick with it. The cd-grid is used pretty sparingly in other implementations so this is a good opportunity to tweak it without too much upset, but I'm not sure when I'll be able to get to it.

left23 (approved):
Looks good to me. Approving.

PR #229, opened 2021-03-17
OPS-7348: revert the auto-incrementing version in API Docs

rupl:
# OPS-7348

I borked the DockerHub image builder and didn't know until now. This script, combined with our weird method of requiring config that uses relative paths, is breaking on account.

Fixing the DockerHub regression caused by 325a7ed


teodorescuserban (approved):
🕹️

PR #233, opened 2021-03-18
HID-2207: post-migration cleanup

rupl:
# HID-2207

Deleting all our migration files and leftover Model files removal of HID Contacts data from prod.

The `showAccount` function was moved inside AuthController without changing anything else about it. It's a safe, non-breaking change since the URL stays put and no other file in the codebase calls it.

Testing Drupal logins will be enough.


orakili (approved):
Working well. Tested login in Drupal sites and editing profile/settings in HID.

PR #235, opened 2021-03-22
HID-2211: add slashes to manually-constructed top-level URLs

rupl:
# HID-2211

Small templating change with big effects on end-users. Their settings pages will now properly link to individual OAuth Client URLs. There was no security problem here because the bug's affect was to link to https://auth.humanitarian.id/someurl.org which 404s on HID and looks ugly, but never sent people to a strange place on the internet.

## Testing

Load your user settings where it lists your authorized OAuth Clients and view source to manually verify that URLs are well-formatted. They were already working fine during local testing which allowed this bug to slip through. Click one or two and confirm that they still link to the intended location.


left23 (approved):
Looks good.

PR #236, opened 2021-03-22
HID-2210: log human-friendly OAuth Client ID when new authorization happens

rupl:
# HID-2210

I was logging the wrong value for `hid.oauth.client_id` in Kibana. Now it will log something like `ocha-gho-local` instead of a 24-char hex string.


orakili (approved):
Looks good to me.

PR #238, opened 2021-03-24
HID-2212 Update common design v3

left23:
Apply updates from the latest common_design releases (currently 4.0.0)
https://github.com/UN-OCHA/common_design/releases


templates/header.ejs
```diff
@@ -41,6 +41,7 @@
     <script async defer nomodule src="/assets/js/css-vars-ponyfill.js"></script>
   </head>
   <body>
+  <div>
```
left23, line 44: This is unfortunate :(

rupl: i'm not sure I understand. Is it something to do with how HID works?

left23: I'm able to remove this extra div now that I've updated the cd.css from common_design
https://github.com/UN-OCHA/common_design/pull/188/commits/8afbd1fa721fd389d9735dcc577ecad578c18e9b

rupl: woohoo! 🎉

left23:
@rupl Feel free to do a casual first review.
Overall it looks good to me and I think this allows for the buttons to be replaced with tags on the Profile page.

The typography from the common_design still needs some work, and I expect in the next week or two as collaboration with the designer is ongoing, this will settle down and hopefully appear more refined.

I'd prefer to wait until we have better typography, rather than make short-term adjustments to HID to compensate, but let me know your preferences. It's most noticeable where there are h2s.

rupl:
I'm also happy to wait. I do think the login page looks a bit strange with the two headings looking visually distinct. I had previously just used the same heading levels and in my mind it wasn't cheating, because it's a homepage serving two main purposes: user login and a "splash" explaining the site's purpose.

Overall it looks great! Nice work. 💯

rupl (approved):
I re-worked the `alert.html` "component" to match the [CD Alert demo](https://demo.commondesign-unocha-org.ahconu.org/demo#cd-alert) as best I could. Otherwise it's great and I merged latest dev to ensure that the CSP work which happened in the interim didn't disrupt anything.

rupl:
I just noticed while tabbing through the site that we didn't have a usable skip-to-content link, despite it being in the markup. I checked prod and this is not a regression, it's a pre-existing bug. The link has `tabindex="-1" plus the wrong classes — also fixed to match CD Demo now.

Whoo boy I will spare you the adventure of discovering that the `<base href="/">` was disrupting the skip-link. What a time I'm having today!

rupl:
Turns out the CSP was indeed preventing fonts from loading. I don't know what I looked at when deciding it was working the first time 😁 

The CSP was brand new work merged in yesterday, so I just fixed that up too since it's an HID server setting.

left23:
@rupl I think this can be merged when you're ready.

PR #239, opened 2021-03-26
HID-2213: E2E scaffolding

rupl:
# HID-2213

Finally! We're automating some of the basic tasks that go into testing HID Auth. This is not yet exhaustive, but I got the itch and wanted to get things started so that future work can incrementally add new tests.

## Testing

There are docs in `CONTRIBUTING.md` and I'd like those to serve as instructions for the PR, so that we can improve them if necessary.


rupl:
Ok I tried not to go _overboard_, but I did manage to do a password reset flow via Mailhog 😎 

We might need the `--runInBand` to avoid having passwords change during other tests such as login, but using the `-t` flag documented in the docs will allow one test to be run without running the others at the same time.

left23:
@rupl Can you confirm the `__env` folder is named as you intended, with the 2 underscores?

I'll update the command in the docs if so 
https://github.com/UN-OCHA/hid_api/blob/687a46463244e549eec1429ddfb13f9f97d627b7/CONTRIBUTING.md#e2e-testing

rupl:
Whoops, let's just go with one underscore. Take the spare one from the dir and add it to the docs ;)

Beware: I _just now_ pushed some more changes underneath you. Sorry about that!

left23 (approved):
I've made one change to the name of the `_env` directory and any references.
I found this PR challenging but after a verbal explanation, I understand how it's working and what the expectations are, and the instructions make more sense in retrospect. I have no suggestions how to clarify this further. I think lack of experience is mainly what was holding me up.
Approving. It's great to have the password reset testing automated.

rupl:
Thanks for the critical review of the docs as well as the functional test.

PR #243, opened 2021-04-16
HID-2218: security headers CSP

rupl:
# HID-2218

Adding headers for Content Security Policy.

Spin up local and make sure the site looks normal. Test if Drupal logins should still work.


orakili (approved):
Tested the HID site and drupal login and everything looks good.

PR #246, opened 2021-04-27
HID-2202: CD Tag labels instead of disabled buttons

rupl:
# HID-2202

Enhancing the ongoing CD v3 work with the new CD Tag.

I had to adjust tag colors to be darker based on a11y feedback from Lighthouse. I feel like long term we should probably raise this to Javi and see if we can adjust the color scheme "from the top" instead of having to balance on-brand with accessible.

The branch also fixes some focus issues. I had to dig back through CD commit history but it seems a bunch of variables were renamed (e.g. `--cd-bright-blue` which effectively removed focus styles from the website, since the value no longer existed.

### DIFF
--- a/assets/css/cd-tag.css
+++ b/assets/css/cd-tag.css
@@ -108,5 +108,5 @@
   --cd-tag--color: #71dbd4;
 }
 .cd-tag--color-6 {
-  --cd-tag--color: #7043ad;
+  --cd-tag--color: #9063cd;
 }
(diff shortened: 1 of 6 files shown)


rupl:
One other thing which I saw but don't feel like fixing: the colors don't work for CD Tag on IE11. I suspect this is due to how I authored the component originally, using CSS Custom Props in a way that the IE ponyfill doesn't support. In stead of trying to hunt the problem down, I just reverted the special colors and all labels are blue now.

left23 (approved):
@rupl Looks good. I agree the blue instead of multi-colour is adequate for this use case.

PR #250, opened 2021-05-05
HID-2225: User and Client model population

rupl:
# HID-2225

This takes care of a couple bugs all related to writing data to a User record. We had seemingly unrelated problems that all ended up being similar once enough data had been collected: trying to update a user with no changes.

So the updates to User.js cover that part. Basically there was a fatal error. You can reproduce by editing your profile and then immediately saving using the "Update Name" button. Having this branch checked out should prevent the error, and it will report having saved your un-edited profile like you'd expect.

The second half stemmed from yet another instance where the `User.authorizedClients` array wasn't getting populated, and additionally we wanted some specially-formatted version of the data on one URL. Instead of manually building it for that page, I moved the code to be a Mongo virtual: essentially, a computed property based on the data itself.

## Testing

1. Log in, edit your Profile, and **without making any change, click <kbd>Update Name</kbd>** — it should show a green success box saying it updated your profile
2. Using API, issue `PUT /api/v3/user/{id}` with an empty payload of `{}` (remember to send your API key as a header) — the response should be the un-edited User object, and additionally the API should no longer crash. Doing this step on the `dev` branch results in API crashing.
3. Log in with a Drupal site and make sure at least one OAuth Client has been Authorized on a user account.
4. Go to User settings and note the presence of a URL on the entry for each Client (first image)
5. Press <kbd>Revoke</kbd> and verify that it does get revoked, and additionally the success message says the name of the site you revoked (second image)

HID-2555-revoke-url

HID-2225-success-name

### DIFF
--- a/api/controllers/ViewController.js
+++ b/api/controllers/ViewController.js
@@ -2,7 +2,6 @@
  * @module ViewController
  * @description Controller for pages that visitors see.
  */
-const URL = require('url');
 const Boom = require('@hapi/boom');
 const Recaptcha = require('recaptcha2');
 const Client = require('../models/Client');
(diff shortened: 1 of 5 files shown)


orakili (approved):
Working well. I could confirm the errors in the `dev` branch and their fix in that PR and all the described tests worked as expected.

PR #251, opened 2021-05-07
HID-2224: always redirect to official URLs for partner sites

rupl:
# HID-2224

Sometimes we show a redirect URL when the partner site had a configuration error. This is a templating change to prevent us from linking anywhere except an official redirect URL that we store inside HID.

## Testing

The easiest way to test this is by falsifying an OAuth Client within HID and leave your Drupal site as-is:

- If necessary, reset the password for a known admin account using the included Mailhog instance.
- Log in as admin, and go to `/admin` to see the list of Clients
- Edit the entry for your chosen OAuth Client, falsifying both the `redirect_uri` and `redirect_urls` fields.
- Try logging in with Drupal and ensure that the value displayed is from the HID DB, and NOT from the Drupal site's `redirectUri` value:

HID-2224-gho-offical

### DIFF
--- a/api/controllers/AuthController.js
+++ b/api/controllers/AuthController.js
@@ -585,8 +585,9 @@ module.exports = {
               },
             );
 
-            // extract hostname from redirect URL
-            errorRedirectUrl = new URL(redirect).origin;
+            // Extract hostname from redirect URL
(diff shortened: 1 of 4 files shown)


orakili (approved):
Works as described.

PR #261, opened 2021-05-12
[HID-1894] fix config

teodorescuserban:
Stop hardcoding things into static files.
We have a new env var named `LOG_LEVEL` which will determine the actual log level in the loggers.
This will be shortly added to the stack for all environments.


config/env/index.js
```diff
@@ -1,5 +1,70 @@
...
+const loggerConsole = new winston.transports.Console();
+
+const loggerLocal = new winston.transports.File({
+  filename: '/var/log/local.log',
+  timestamp: true,
+});
+
+const transports = [];
+
+if (env == 'local') {
```
rupl, line 33:
```suggestion
if (env === 'local') {
```

rupl (approved):
inline suggestion but otherwise looks good. still working on my local. Tested all the major DB things: login to HID, login _with_ HID from somewhere else, profile edits, revoke client, edit client as admin.

config/env/index.js
teodorescuserban: yeah eslint complained about this as well. fix incoming.

PR #263, opened 2021-05-13
no more hardcoded authorizedDomains! yey!

teodorescuserban:
# HID-1894

Moving our hardcoded domains to an environment variable.


rupl (approved):
love it!

PR #267, opened 2021-06-01
fix environment use in commands

s0:
Changes were made to environment in #261 but usage was not updated in
commands. This commit makes that update.

cc @teodorescuserban @rupl


rupl (approved):
Whoopsie! Yup that would have caused errors on prod inside Jenkins. We don't run cron in dev and maybe we should, in order to surface problems with these commands. Thanks for the PR!

PR #271, opened 2021-06-14
HID-2233: ranomize new  OAuth Client IDs

rupl:
# HID-2233

Preventing a rare potential error before it bites us. Randomizes the `client_id` of a new OAuth Client since we generate them via Jenkins. Now you can run the job multiple times (say you want dev/stage/live) and it is quite unlikely to create collisions.

## Testing

Run this command inside the `api` container a couple times in a row:

```sh
docker-compose exec api node commands/createOAuthClient.js
```

Old behavior would have had errors on the second time and afterwards. PR should be able to run it essentially unlimited number of times. 308,915,776 times to be exact.

### DIFF
--- a/commands/createOAuthClient.js
+++ b/commands/createOAuthClient.js
@@ -17,15 +17,18 @@ mongoose.connect(store.uri, store.options);
 const Client = require('../api/models/Client');
 
 // Generate a random secret each time we run the command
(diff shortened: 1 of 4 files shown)


orakili (approved):
Looks good. If we wanted to go further and ensure there can never be a collision, we could check if the generated ID already exists and regenerate a new one until it's unique.

rupl:
I did think of doing fancier things, but the randomization code was already sitting there so I reused it. ¯\\\_(ツ)\_/¯

PR #273, opened 2021-06-16
HID-1997: account.json error

rupl:
# HID-1997

We used to serve HTTP 200 with an empty object `{}` when the Authorization header was missing or invalid. Then at some point the API started serving HTTP 500, I would guess when we changed the payload that gets sent for v3.

This PR brings us in line with the [official OpenID Connect spec](https://openid.net/specs/openid-connect-core-1_0.html#UserInfoError) by returning HTTP 401 for both missing and invalid tokens. I also updated the [appropriate wiki section](https://github.com/UN-OCHA/hid_api/wiki/Integrating-with-HID-via-OAuth#step-3--request-user-account-info) with an example.

## Testing

Visiting `/account.json` in your browser should now produce HTTP 401 instead of 200 or 500. Sending the Authorization header, or logging in with Drupal should (naturally) produce the user account.

The other stuff that got shuffled around is related to 2FA, and there are no actual changes, only moving code within the file. Testing something like a password reset on a 2FA user is sufficient to ensure it's working.

### DIFF
--- a/_tests/e2e/_env/example.js
+++ b/_tests/e2e/_env/example.js
@@ -1,10 +1,6 @@
 module.exports = {
   baseUrl: 'http://hid.test',
 
-  // Test params
-  timeoutLong: 15000,
-  timeoutLonger: 30000,
-
   // Users/data
   testUserId: '',
   testUserNameGiven: 'Test',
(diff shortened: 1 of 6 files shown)


orakili (approved):
Worked well.

PR #274, opened 2021-06-17
HID-2230: typo in user-facing message

rupl:
# HID-2230

The actual issue in the ticket isn't resolved, but I thought I'd slide this in before leaving for holiday.


orakili (approved):
👍

PR #275, opened 2021-08-10
HID-2248: better errors and logging for password resets

rupl:
# HID-2248

We got a report that someone saw what looks like a 500 error during password reset. It's just uninformative, and on top of that we weren't logging anything when there was a real error.

E2E passed on my local. To test you can do two things:

- Visit http://auth.hid.test/new_password - should show a red error with a link to reset your password
- Successfully reset your password after clicking the link. Mailhog is at http://localhost:8025

### DIFF
--- a/api/controllers/ViewController.js
+++ b/api/controllers/ViewController.js
@@ -361,7 +361,10 @@ module.exports = {
   async newPassword(request, reply) {
     request.yar.reset();
     request.yar.set('session', {
-      hash: request.query.hash, id: request.query.id, time: request.query.time, totp: false,
+      hash: request.query.hash,
+      id: request.query.id,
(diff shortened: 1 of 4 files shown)


left23 (approved):
I can confirm I get an error message when I go to http://hid.test/new_password (my local) and I can successfully reset my password after clicking the link.

@rupl My local set up still works but note the different URL than what is in the PR description.
http://auth.hid.test/new_password vs http://hid.test/password
I _think_ this doesn't matter but I'm rusty so let me know I've missed something.

rupl:
Whoops, yeah all those URLs are valid, especially for local development. If it works, it works 👍 

thanks for reviewing!

PR #277, opened 2021-08-13
HID-2250

rupl:
# HID-2250

Making an error message more clear, plus logging. I also added the ability to output an "Error type" within an error message, to help the Content Squad field support requests. On any error (or any message at all) you can now include a `error_type` property, and it will manually format the string, plus output a timestamp from the server. Combined with wording on the error message, we'll be able to provide Content Squad (and our own log spelunking) better data to go hunt down problems:

HID-2250-error-code-5

## Testing

Try to register a new account (even though reCAPTCHA is incorrectly configured). It should show this message.


left23 (approved):
I confirm I get the error message when I try to register a new account.

PR #279, opened 2021-08-17
HID-2064: split email confirmation into two routes

rupl:
# HID-2064

Marking as a draft until the open PRs are merged.

I split the route into two pieces and since we're incrementing the major version on next deploy, I took the liberty of shoving all the breaking changes into one branch.

The following endpoint is gone:
- `PUT /api/v3/user/emails/{email}`

In its place are two new ones:
- `POST /api/v3/user/emails/{email}`
- `POST /api/v3/user/emails/validate`

Not only was `PUT` inaccurate (since the email message it triggers is unique each time) but the code no longer serves two functions based on the payload you send.

## Testing

- Use API to create an email, and copy data out of Mailhog to craft the second API request.
- Use website to create an email, and click the email link to confirm an address.

Brand new E2E awaiting you, hopefully to reduce manual clicking and stuff.

```
npm run e2e -- -t "ProfileEmails"
```

## Known issues

- HID-2200: It sends "welcome to HID" emails if you use the API to reconfirm your primary. The Auth website doesn't provide that possibility so it's not common in real life.
- HID-2254: the hash does not get invalidated after the first-use. If you click the email link again, it behaves as if you never did it before, which compounds the effect of the first bug (one "welcome to HID" email for each time you click the link).

### DIFF
--- a/_tests/e2e/Login.test.js
+++ b/_tests/e2e/Login.test.js
@@ -1,4 +1,5 @@
 import env from './_env';
+import utils from './_utils';
 
 describe('Login', () => {
   let page;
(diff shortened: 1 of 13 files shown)


orakili (approved):
Tested and working as advertised.

rupl:
Thanks!

PR #280, opened 2021-08-24
HID-2219: password resets + email confirmation via recovery email address

rupl:
# HID-2219

- Splits the generate/validate functions related to hashes. You no longer have to do `user.validHash(something, 'password_reset', second_arg)` and instead each one has a dedicated name: `user.validHashPassword()` 
- Incorporates the `emailId` parameter into `generateHashPassword`/`validHashPassword` so that a recovery `emailId` can be passed in and considered confirmed once the password reset is totally successful. If the password fails to reset for any reason, the email will also remain unconfirmed.
- Removes `reset_url` as a param from `POST /api/v3/user/password-email` — this is no longer a decoupled API/app so it's totally useless.
- Adds an optional `emailId` payload property to `POST /api/v3/user/password` — it will become required when we complete the deployment transition window to formally invalidate the old reset link structure.

## Auth-specific

- Duplicates our `/new_password` route to `/new-password` — this was the last route with an underscore. In a future deploy we will remove the route entirely. For now, leaving it in place means that the 24-hour window of time after deployment will not break the password reset links of users who generated them beforehand.

## Testing

There is new E2E to support the new recovery email address capability when you have this branch checked out from the beginning. However, a simulation of the upcoming deployment will be necessary:

1. Checkout the `dev` branch and initiate a password reset to the **primary** email address (since recovery resets aren't allowed in `dev` until this PR is merged)
1. Note the message in mailhog, and then checkout `HID-2219` branch
1. Once the site is restarted on the new code, continue in Mailhog, using the password reset generated from the `dev` branch.
1. Unless I've overlooked something, the old route and hash format should continue working.
1. Now initiate a password reset from the primary again
1. Click the mailhog link and confirm that it works too (the reset link has a new `emailId` param in it)
1. Finally, add a recovery email to an account. Ignore the "confirm this email" message.
1. Initiate a password reset to the unconfirmed recovery address.
1. The password link should also function, allowing you to reset the password.

⚠️  If you generate multiple reset links without first using the older ones, the newer ones get invalidated as you reset your password. This is by design, since the user's hashed password is part of the input for `generateHashPassword()`. Following these exact steps should yield success at each step.


api/controllers/UserController.js
```diff
@@ -1506,8 +1466,25 @@ module.exports = {
       user = await AuthPolicy.isTOTPValid(user, token);
     }
 
+    // If an emailId parameter was present, we will use that value. However, if
+    // nothing was sent, we need to determine the ID of the user's primary email
+    // address of the as a default value.
```
orakili, line 1471: > address of the as a default value

Some words seem to be missing.

orakili (approved):
Working well. Tested the following:

1. Password reset with old code with primary email address
2. Checked that password reset from old code still worked with new code
3. Password reset with new code with primary email address (confirming presence of emailId param)
4. Password reset with new code with **confirmed** secondary email address (confirming presence of emailId param)
5. Password reset with new code with **unconfirmed** secondary email address (confirming presence of emailId param)
6. Tested password reset link with new code when removing emailId for primary or secondary emails and in both cases, I get the expected link not valid message.

api/controllers/UserController.js
rupl: I left that in there to see if you're paying attention 😉 

thanks for the spot!

PR #281, opened 2021-08-24
HID-2150: avoid sending email_alert messages to unconfirmed recovery addresses

rupl:
# HID-2150

Previously, when you added a recovery email to your account, every email on your profile got a message notifying you about it. The PR prevents unconfirmed addresses from getting such messages.

## Testing

1. On your local, delete all the recovery emails on a test account of your choosing, so that only the primary remains.
2. Now, add a recovery email to that account. Auth or API calls are both fine. Do NOT confirm it by clicking link in Mailhog.
3. Add another recovery email (skip confirmation)
4. Add another recovery email (skip confirmation)
5. Check Mailhog and see that only _confirmed_ addresses are getting emails with subject line "Security alert for your HID profile" — it should be one per recovery email which will be mentioned in the message body.

### DIFF
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -948,13 +948,14 @@ module.exports = {
     // Send confirmation email
     await EmailService.sendValidationEmail(user, emailToAdd, savedEmailId);
 
-    // If the email sent without error, notify the other emails on this account.
+    // If the email sent without error, notify the other confirmed emails on
+    // this account.
     const promises = [];
     for (let i = 0; i < user.emails.length; i++) {
-      // TODO: probably shouldn't send notices to unconfirmed email addresses.
-      //
-      // @see HID-2150
-      promises.push(EmailService.sendEmailAlert(user, user.emails[i].email, emailToAdd));
+      const thisEmail = user.emails[i];
+      if (thisEmail.validated) {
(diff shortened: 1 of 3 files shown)


api/controllers/UserController.js
```diff
@@ -914,81 +892,77 @@ module.exports = {
...
-    for (let i = 0; i < record.emails.length; i++) {
-      // TODO: probably shouldn't send notices to unconfirmed email addresses.
-      //
-      // @see HID-2150
-      promises.push(EmailService.sendEmailAlert(record, record.emails[i].email, email));
+    for (let i = 0; i < user.emails.length; i++) {
+      const thisEmail = user.emails[i];
+      if (thisEmail.validated) {
+        promises.push(EmailService.sendEmailAlert(user, thisEmail.email, emailToAdd));
+      }
```
rupl, line 958: This is the core fix to the ticket. There were lots of other standardizations I made (e.g. changing `record` to `user` and renaming the very easy-to-confuse `email` variable to be more descriptive).

rupl:
@orakili I have un-drafted this PR while intentionally maintaining the base branch `HID-2219`. I'll probably merge them all together when I get back next week, using the dev deploy as "practice" for the eventual prod deploy.

PR #283, opened 2021-08-31
HID-2255: server-side validation of password reset, plus restore client-side validation

rupl:
# HID-2255

This started out as a small fix for the CSP, which disabled our client-side validation.

Since that happened, I'd actually noticed some intermittent errors during password resets and never could figure out what it was. It ends up that we weren't enforcing the `confirm_password` field on the server, which combined with lack of client-side validation, meant it was possible to change your password to something unexpected.

PR fixes both problems: we now have both server- and client-side validation, instead of nothing at all 🙈 

This also conveniently gets included in the 4.0.0 release, since the `confirm_password` property of the payload is now required.

## Testing

Basically, just reset a user's password and fill the form out incorrectly in various ways. It should now stop you from doing anything except the right thing.

To test server-side validation of form submissions, you can comment out the client-side JS file on your local, or disable JS in the browser.

### DIFF
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -1551,7 +1551,7 @@ module.exports = {
     // Update user in DB.
     await user.save().then(() => {
       logger.info(
-        `[UserController->resetPassword] Password updated successfully for user ${user.id}`,
+       '[UserController->resetPassword] Password updated successfully',
         {
           request,
           security: true,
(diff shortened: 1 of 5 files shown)


left23 (approved):
@rupl Approving. I tested with multiple incorrect passwords, and with JS disabled, and this came up each time:
Selection_005
I also tested with the correct password criteria and I was able to reset.

PR #284, opened 2021-09-01
HID-2261: remove isSuccess variable from message.html

rupl:
# HID-2216

At some point I changed a conditional from checking truthiness, to checking if it existed at all. That was a no-no, and caused messages to show up when they shouldn't. In this case, when you deleted your account, the website said you can now login to HID.

After removing the instances of `isSuccess: false` which was causing it to show up _specifically when it shouldn't_, I realized there is one instance of the template left. That's after password reset. I think people will know that they can login after resetting their password, so I removed the whole thing.

## Testing

Reset the password of a random account, login with it, and delete the account. At each step, the user feedback should function, and no templating errors should occur.

### DIFF
--- a/api/controllers/AuthController.js
+++ b/api/controllers/AuthController.js
@@ -671,7 +671,6 @@ module.exports = {
             ${linktoPartnerSite || ''}
           `,
         },
-        isSuccess: false,
       });
     }
   },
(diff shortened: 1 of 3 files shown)


left23 (approved):
@rupl Approving.

Feedback is clear and I can confirm there is no extra or confusing messages after account deletion.
Selection_007

PR #285, opened 2021-09-06
HID-2262: adopt and simplify UserPolicy.canUpdate

rupl:
# HID-2262

We restricted the API a little too much when we removed searching capabilities from HID Contacts. This PR restores any user's permission to GET themselves from the API.

## Testing

1. Request `GET /api/v3/user/{id}` from a regular, non-admin user. It should return the user's object as long as you send your Bearer token.
2. Change the value of `{id}` to some other value without changing the number of characters (i.e. replace any `1` with a `2`) and it should respond with 403 Forbidden.
3. Use an admin account to search for any individual user ID. All user IDs, including your own, should work.

### DIFF
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -323,7 +323,7 @@ module.exports = {
    *   '404':
    *     description: Requested user not found.
    */
-  async findOne(request, reply) {
+  async findOne(request) {
     const user = await User.findById(request.params.id);
 
     // If we found a user, sanitize and return it
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -331,7 +331,7 @@ module.exports = {
     // If we found a user, sanitize and return it
     if (user) {
       logger.info(
-        '[UserController->find] Displaying one user by ID',
+        '[UserController->findOne] Displaying one user by ID',
         {
           request,
           user: {
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -240,77 +240,7 @@ module.exports = {
    *   '401':
    *     description: Requesting user lacks permission to query users.
    */
-
-  /*
-   * @api [get] /user/{id}
-   * tags:
-   *  - user
-   * summary: Returns a User by ID.
-   * parameters:
-   *   - name: id
-   *     description: A 24-character alphanumeric User ID
-   *     in: path
-   *     required: true
-   *     default: ''
-   * responses:
-   *   '200':
-   *     description: The requested user
-   *     content:
-   *       application/json:
-   *         schema:
-   *           $ref: '#/components/schemas/User'
-   *   '400':
-   *     description: Bad request.
-   *   '401':
-   *     description: Requesting user lacks permission to view requested user.
-   *   '404':
-   *     description: Requested user not found.
-   */
   async find(request, reply) {
-    if (request.params.id) {
-      const criteria = { _id: request.params.id };
-
-      // Do not show user if it is hidden
-      if (
-        !request.auth.credentials.is_admin
-        && request.auth.credentials._id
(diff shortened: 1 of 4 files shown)


api/policies/UserPolicy.js
```diff
@@ -92,7 +55,32 @@ module.exports = {
...
+    //   OR any user is targeting themselves
+    // THEN are allowed to update the target user.
+    if (request.auth.credentials.is_admin || request.auth.credentials.id === request.params.id) {
+      const user = await User.findById(request.params.id);
+      if (!user) {
+        logger.warn(
+          `[UserPolicy->canUpdate] User ${request.params.id} not found`,
+        );
+        throw Boom.notFound();
+      }
```
orakili, line 69: This breaks the [`/api/v3/user` endpoint (GET)](https://github.com/UN-OCHA/hid_api/blob/9fd38d6ca44d974c17bab359d525110cb6556ffd/config/routes.js#L376): we get a 404 (because no user id) instead of the full list of users when accessed with an admin user (we get a 403 as expected for a non admin user).

Maybe we can still use the `canFind` user policy for this endpoint and if there is an `id` provided return the result of `canUpdate`, or explicitly separate the `/api/v3/user` and `/api/v3/user/{id}` endpoints and use `canFind` for the former and `canUpdate` for the latter. What do you think?

orakili (changes requested):

api/policies/UserPolicy.js
rupl: Thanks, can't believe I forgot to re-test that. Yes, I will simply split the route definition into two, so we can keep the old policy when no ID is submitted.

orakili (approved):
The API is returning what is expected for normal and admin users. 

However I find that the log messages are potentially a bit confusing due to the use of `canUpdate` for the GET endpoints.

For example, when requesting `/api/v3/user/{id}` **as an admin** with an ID that doesn't match any user, then the log message is the "canUpdate] User xxx not found" instead of the `find] Could not find user xxx" from the `findOne` method.

I think it could be clearer to use `canFind` for both GET endpoints and replace:

```
if (request.auth.credentials.is_admin) {
```

with 

```
if (request.auth.credentials.is_admin || request.auth.credentials.id === request.params.id) {
```

That covers the admin only access to `/user` and the access  to `/user/{id}` to admins and the requesting user if it's the same ID.

api/controllers/UserController.js
```diff
@@ -368,6 +298,64 @@ module.exports = {
...
+        },
+      );
+
+      user.sanitize(request.auth.credentials);
+      return user;
+    }
+
+    // Finally: if we didn't find a user, send a 404.
+    logger.warn(
+      `[UserController->find] Could not find user ${request.params.id}`,
```
orakili, line 350:
```suggestion
      `[UserController->findOne] Could not find user ${request.params.id}`,
```

api/controllers/UserController.js
```diff
@@ -368,6 +298,64 @@ module.exports = {
...
+   *   '404':
+   *     description: Requested user not found.
+   */
+  async findOne(request) {
+    const user = await User.findById(request.params.id);
+
+    // If we found a user, sanitize and return it
+    if (user) {
+      logger.info(
+        '[UserController->find] Displaying one user by ID',
```
orakili, line 334:
```suggestion
        '[UserController->findOne] Displaying one user by ID',
```

rupl:
ok I have incorporated the feedback and verified all of the following:

- use `canFind` for any `GET` method
- if a 404 is to be output, allow that to be logged inside `UserController` instead of `UserPolicy`
- if an operation should be prevented on the grounds of permissions, the error is logged by `UserPolicy`
- both `find`/`findOne` should log correct methods (I committed your suggested updates)

orakili (approved):
👍

PR #287, opened 2021-09-16
HPC-7918: remove deprecated fields from account.json

rupl:
# [HPC-7918](https://humanitarian.atlassian.net/browse/HPC-7918)

Final cleanup for HIDv3 migration. Removes the following fields from `account.json` so that we can complete our DB pruning:

- `given_name`
- `family_name`
- `user_id`


orakili (approved):
Looks good to me.

PR #291, opened 2021-09-17
HID-2260: implement session storage backend using Redis

rupl:
# HID-2260

We found that HID has an implicit dependency on the load balancer cookies. It turns out that the session management library, `yar`, allows us to store session data up to a certain amount, 1KB. Once you exceed that limit, it starts dumping the data to a server cache, which uses container memory as its default backend.

If the load balancer stickiness gets disabled, OR if we had to scale up HID API containers due to heavy load, the user's session data has a high chance of appearing to not exist. The UX of this bug is ugly, resulting in people seeing a login form after already being logged in. Or sometimes, it looked like the login form wasn't working when the cookie limit got exceeded during login itself.

The PR enables a Redis backend for the session storage, so that all API containers can share storage, pushing and pulling data as needed. I've reconfigured the storage to be used in all cases, by setting a storage limit of 0 on the cookies themselves. I figure it's more consistent this way, and we only have one approach to debug.

Aside from Redis I added a whole lot of logs to see when pages get rendered, and also simplified the conditionals for cookies themselves, which I'd brainlessly pasted all over the place as time went on.

While we were touching session behavior, I extended the cookie lifetime from 4 hours to 1 week. I feel this is a nice UX improvement and there's no OICT policy against it that I know of.

## Testing

The PR is in draft because **the HID stack needs to be pulled to latest `main` branch, which has Redis configured** in order to test this properly. Additionally, I was told to change `env/common/etc/nginx/http.d/vhost_api.conf` to disable IP hashing, plus change the internal hostname to `api` in order to simulate the load balancer:

```diff
diff --git env/common/etc/nginx/http.d/vhost_api.conf env/common/etc/nginx/http.d/vhost_api.conf
index b2dd0cf..731083b 100644
--- env/common/etc/nginx/http.d/vhost_api.conf
+++ env/common/etc/nginx/http.d/vhost_api.conf
@@ -1,12 +1,12 @@

 upstream hid-api-backend {
-    server hid-api:3000;
-    ip_hash;
+    server api:3000;
+#    ip_hash;
 }
```

Once the stack is available, here is the procedure:

1. When starting your HID stack, use `--scale api=3` as an argument, like so: `docker-compose up -d --scale api=3` — this will create enough API containers to reproduce the bug easily.
2. Check out `dev` branch
3. Log in with a non-2FA user, navigate to `/settings/security/` and click the button to enable 2FA
4. If not the first try, after a few attempts of clicking "Enable 2FA" you should observe that the login form is displayed, instead of the 2FA activation form. This is due to the cookie data exceeding the limit, being written to container memory, which doesn't exist when a different API container serves the following page load.
5. Now check this branch out.
6. Repeat steps 3-4. You should be able to repeatedly load the 2FA form. You can simply click "Enable 2FA" and without completing the process, click the "Security" tab again, "Enable 2FA" again, etc. You should see the QR code every time, and **never** see the login form unexpectedly.
7. If you want to truly verify, open your browser devtools and click the "Enable 2FA" button. Now copy the `POST` request in the devtools as `cURL` (you can do this in Chrome, FF, or Safari). Execute the `cURL` command in your terminal.
8. Now log into the redis container `docker-compose exec redis redis-cli`
9. On the redis CLI, run `keys *` and from there run `get [KEY]` to see the data being stored. It should contain a base64-encoded PNG which represents the QR code.

### DIFF
--- a/api/controllers/ViewController.js
+++ b/api/controllers/ViewController.js
@@ -635,7 +635,7 @@ module.exports = {
     const cookie = request.yar.get('session');
     if (!cookie || (cookie && !cookie.userId) || (cookie && !cookie.totp)) {
       logger.info(
-        '[ViewController->user] User tried to view dashboard without session. Redirecting to login.',
+        '[ViewController->user] User had no session. Redirecting to login.',
         {
           request,
         },
@@ -679,6 +679,12 @@ module.exports = {
     // If the user is not authenticated, redirect to the login page
     const cookie = request.yar.get('session');
     if (!cookie || (cookie && !cookie.userId) || (cookie && !cookie.totp)) {
+      logger.info(
+        '[ViewController->profile] User had no session. Redirecting to login.',
+        {
+          request,
+        },
+      );
       return reply.redirect('/');
     }
 
@@ -706,6 +712,12 @@ module.exports = {
     // If the user is not authenticated, redirect to the login page
     const cookie = request.yar.get('session');
     if (!cookie || (cookie && !cookie.userId) || (cookie && !cookie.totp)) {
+      logger.info(
+        '[ViewController->profileEdit] User had no session. Redirecting to login.',
+        {
+          request,
+        },
+      );
       return reply.redirect('/');
     }
 
@@ -734,6 +746,12 @@ module.exports = {
     // If the user is not authenticated, redirect to the login page
     const cookie = request.yar.get('session');
     if (!cookie || (cookie && !cookie.userId) || (cookie && !cookie.totp)) {
+      logger.info(
+        '[ViewController->profileEditSubmit] User had no session. Redirecting to login.',
+        {
+          request,
+        },
+      );
       return reply.redirect('/');
     }
 
@@ -829,6 +847,12 @@ module.exports = {
     // If the user is not authenticated, redirect to the login page
     const cookie = request.yar.get('session');
     if (!cookie || (cookie && !cookie.userId) || (cookie && !cookie.totp)) {
+      logger.info(
+        '[ViewController->profileEmailsSubmit] User had no session. Redirecting to login.',
+        {
+          request,
+        },
+      );
       return reply.redirect('/');
     }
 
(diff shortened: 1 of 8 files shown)


rupl:
@orakili I believe all the pieces are now in place. You will need to update your HID stack and recreate all containers to test this with redis. Ping me if you have any questions about setup.

rupl:
whoooops nope, this is not ready. I created 2FA bypass by "simplifying" the conditionals on all the routes. Back to me.

rupl:
ok luckily that was quick to fix. I didn't revert it, but restored all the 2FA conditionals.

api/controllers/ViewController.js
```diff
@@ -593,10 +648,42 @@ module.exports = {
...
     // If the user is not authenticated, redirect to the login page
     const cookie = request.yar.get('session');
-    if (!cookie || (cookie && !cookie.userId) || (cookie && !cookie.totp)) {
+    if (!cookie || !cookie.userId || !cookie.totp) {
+      logger.info(
+        '[ViewController->user] User had no session. Redirecting to login.',
+        {
+          request,
+        },
+      );
```
teodorescuserban, line 657: DRY? Maybe calling `noSessionRedirectsToLogin` ? :)

rupl: I agree with you and tried to avoid it. Problem is, I had difficulty using `reply` to act on the conditional if it's inside another function. So basically it only removes part of the boilerplate but you're still left with:

```js
if (!hasValidSession(request, 'METHOD_NAME')) {
  return reply.redirect('/');
}

// Now collect cookie data AGAIN so this function can use it.
const cookie = request.yar.get('session');

... stuff ...
```

And this results in some "HID Classic" lie-prone logging, where the actual log entry appears BEFORE you redirect, and it would be possible to not even do the redirect, like so:

```js
if (!hasValidSession(request, 'METHOD_NAME')) {
  // I'm not redirecting, but `hasValidSession` already created a log entry saying I was gonna do that
  return reply.view('error');
}
```

Additionally, while tinkering I couldn't find a nice way to make the logs record which function was punting the user back to login. I tried looking at automatic stuff like `args.caller` or even a separate param on the function but I could never find the "magic" I wanted. So the examples above have `'METHOD_NAME` as a required param in the hypothetical `hasValidSession()` function, so that they can log on behalf of said method. Again, kinda backwards.

Anyway, that's why I copy/pasted the code into each method. I will gladly take suggestions if all the above issues can be considered!

teodorescuserban: Yeah, it is probably not worth it all things considered.

PR #292, opened 2021-09-17
HID-2216: implement one remediation measure against session fixation

rupl:
# HID-2216

⚠️  This branch uses #291 as its base. It will be in Draft until that PR is merged ⚠️ 

This PR attempts to protect against some types of session hijacking, by resetting the session once a user successfully logs in. The result is that even if you stole a user's cookie during login, it won't be any good by the time they submit the form and finish the login process. They'll have a new ID and the old one won't exist.

## Testing

I added a new "command" to make testing this easier. It uses the stack environment to decrypt cookies that you receive in your browser.

1. Load HID login form in browser. Copy the `session` cookie from devtools.
2. `docker-compose exec api node commands/readCookie.js --cookie="__PASTE__"` — note the session ID that gets output.
3. Now login using the HID form. Once the page shows your user details, copy the `session` cookie again.
4. `docker-compose exec api node commands/readCookie.js --cookie="__PASTE__"` — it **MUST** be different than in Step 2.

### DIFF
--- a/api/controllers/AuthController.js
+++ b/api/controllers/AuthController.js
@@ -266,7 +266,7 @@ module.exports = {
    * another website.
    */
   async login(request, reply) {
-    // Grab cookie
+    // Get user session.
     const cookie = request.yar.get('session');
 
     // If the user has a User ID set, and passed the optional 2FA challenge,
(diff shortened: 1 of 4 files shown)


orakili (approved):
Works as described.

PR #293, opened 2021-09-21
HID-2192: remove special cases from /account.json

rupl:
# HID-2192

We asked around the team and also checked the logs. The affected variable which everything was using does not seem to ever be present in our requests nowadays.

Nevertheless I marked it as a breaking change and it will bump our API's major version. I would have liked to group it into the current release but oh well.

As a bonus, I looked into the token-to-user stuff and got the OAuth Client ID coming through via `request.auth.artifacts` which is now logged on each request to `/account.json` when you use a short-lived OAuth token.

# Testing

- **Make sure Drupal login still works.** It should log the OAuth `client_id` of the website when it logs "[AuthController->showAccount] calling /account.json for ..."
- **Make sure `GET /account.json` still works with your personal JWT** (this is technically not supposed to happen but no need to crash if it does)

### DIFF
--- a/api/controllers/AuthController.js
+++ b/api/controllers/AuthController.js
@@ -1005,26 +1005,6 @@ module.exports = {
       },
     );
 
-    // Special cases for legacy compat.
-    //
-    // @TODO: in testing this, it seems that the `currentClient` param is not
-    //        present when this function runs. Investigate whether we need these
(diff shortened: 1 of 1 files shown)


orakili (approved):
Working as expected.

PR #296, opened 2021-09-22
HID-2276: store historical password hashes and prevent re-use

rupl:
# HID-2276

During our annual security self-assessment for OICT, this stood out as the only relevant item we don't implement. Their guidance says that users should not reuse the previous five passwords.

## Testing

- Changing password via API call should behave almost identically, but you can no longer switch between two passwords like you used to be able to. Passwords can no longer be reused and the error code remains the same, but the message says "New password must be different than previous passwords" — the old response used to be "New password is the same as old password."
- While logged-in, using the password change form at `/settings/password` will tell you that a password cannot be reused (same error text as API response)
- If you try to re-use passwords when resetting passwords via email links, it will show the less specific failure notification informing you that the password could not be changed. The new log entry will report: "[UserController->resetPassword] Could not reset password. New password must be different than previous passwords."

### DIFF
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -573,7 +573,7 @@ module.exports = {
     }
 
     // Look up user in DB.
-    const user = await User.findOne({ _id: userId });
+    const user = await User.findById(userId);
 
     // Was the user parameter supplied?
     if (!user) {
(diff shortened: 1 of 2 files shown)


orakili (approved):
Tested and working well.

I indeed had to generate 5 new passwords before being able to re-use the one before those 5 new passwords.

PR #298, opened 2021-09-24
HID-2273: upgrade @hapi/yar

rupl:
# HID-2273

Update our dependency for session management. Had a trivial breaking change, that the plugin name is now namespaced with `@hapi/`. They also bumped node.js minimum but we're still ahead of their requirements at this time.

## Testing

- Drupal login + normal login should be enough. The site broke pretty spectacularly before I found that plugin name, and afterward it managed sessions like usual, including the new redis integration.


orakili (approved):
👍

PR #303, opened 2021-10-12
[OPS-7809] migrate to ECR

teodorescuserban:


.github/workflows/build.yml
```diff
@@ -0,0 +1,55 @@
+name: Create image
+
+on:
+  push:
+    branches:
+      - dev
+      - main
+      - master
+      - ops-7908
```
rupl, line 9:
```suggestion
      - OPS-*
```

Why not future proof this instead of only building this one ops-related branch?

PR #304, opened 2021-10-13
HID-2283: prevent errors due to case-sensitivity

rupl:
# HID-2283

We have users reporting errors when their email addresses contain capital letters anywhere in the address. This is because we store email addresses in DB all lowercase, and we weren't lowercasing them when doing password reset lookups.

Additionally, I noticed that emails aren't actually getting verified when password resets are successful, which can effectively lock someone out of their account permanently if they only have a primary email and never click the initial confirmation link. I got myself into this situation while testing locally with a fresh account.

While looking at this I removed the old `/new_password` routes which will only be used by emails that are a month old by the time we deploy to prod. I also renamed the template to `new-password.html` so that the whole directory is consistent with uses of dashes instead of underscores.

# Testing

1. Register new account with at least one CAPITAL letter in the email address `CAPITALS@example.com`
2. **Ignore the confirmation email**
3. Initiate a password reset by typing your new email with at least one capital letter. Note: it does NOT need to match what you typed when you registered. This is basically the bug we're fixing. e.g. `CaPiTaLs@example.com`
4. You should NOT see an error, and SHOULD see a message saying "if [email] exists, we sent a password reset link"
5. After resetting password, you should be able to login with your new password (despite having ignored the initial email confirmation link).

### DIFF
--- a/api/models/User.js
+++ b/api/models/User.js
@@ -624,7 +624,7 @@ UserSchema.methods = {
   emailIndex(email) {
     let index = -1;
     for (let i = 0, len = this.emails.length; i < len; i++) {
-      if (this.emails[i].email === email) {
+      if (this.emails[i].email === email.toLowerCase()) {
         index = i;
       }
     }


rupl:
I've added one more commit that I believe would fix all instances of this error. I reversed some of the changes in this branch locally, added the `toLowerCase()` to the `User.emailIndex()` function, and they all seem to work.

Lowercasing during the lookup ensures that it always normalizes no matter what input is provided (since this can happen at many points). This will also fix a new bug I noticed while testing, which would show an error-colored success message when adding an email address with capitals directly to your logged-in profile page:

HID-2283-profile-email-1

HID-2283-profile-email-2

rupl:
> The local-part of a mailbox MUST BE treated as case sensitive.

On second reading, this is about SMTP protocol, not about providing/storing email addresses themselves. In my interpretation, someone can honor this requirement implementing SMTP while also normalizing storage of email addresses for their own reasons (like HID, Google, et. al seem to be doing).

Come to think of it, I feel like case-insensitive email addresses is a security feature. Imagine if someone came by and grabbed your gmail, but with one capital letter. That would be a real mess if it were easy and commonplace.

orakili (approved):
Works as described. Tested a few other actions with different versions of the email address and  they worked as well.

PR #305, opened 2021-10-13
HID-2286: offer password visibility toggle buttons

rupl:
# HID-2286

Our first wave of signups from RW resulted in many requests to toggle visibility of password fields, so that users registering can visually confirm their input. The branch creates such a mechanism which uses JS to toggle the input's `type` attribute between `password`/`text`, then applies it to all password fields.

This branch is client-side behavior only. The transmission of data hasn't changed. The names of inputs have not been changed. The only thing that changes is the input's `type` and that doesn't affect validation or submission payload.

## Testing

Each of the activities listed above should be tested for the frontend functionality:

- Sign up with a new account
- Reset the password of an existing account
- Login with an existing account
- Change password while logged in

I personally tested with:
- macOS + Chrome/FF (keyboard/mouse)
- iOS 15 + Safari (touch)
- Win10 + Edge/IE11 on Browserstack (keyboard/mouse)

Any other combos would be welcome as well.

@left23 given our time difference, I've deployed this branch to dev so that you can inspect the work without bothering to setup your local. I hope that speeds review up for you! https://dev.auth-humanitarian-id.ahconu.org/


rupl:
Thanks for the review!

PR #306, opened 2021-10-14
HID-2280: permanently disable password expiry

rupl:
# HID-2280

This branch permanently removes the concept of password expiry:

- User properties that track password expiry reminders
- Code to update such properties
- Email templates used for password expiry reminders

Additionally we fixed a flaw which was creating work for the Content Squad in the form of support requests

- Updated code to record when logged-in users voluntarily update passwords. Previously you could update a password voluntarily, and the system would continue sending password reminders as if you didn't make any change.

# Testing

Check the branch out and run the migration:

```sh
# Do this from the stack repo inside env/local
$ docker-compose exec api node commands/migratePasswordExpiry.js
{"migration":true,"level":"info","message":"[commands->migratePasswordExpiry] Removed passwordResetAlert30days, passwordResetAlert7days, passwordResetAlert fields from all users","request":{"path":{},"query":{},"payload":{},"headers":{},"auth":{}},"@timestamp":"2021-10-14T09:18:24.497Z","user":{},"oauth":{}}

# Run this from same directory
> db.user.count({passwordResetAlert30days: {$exists: true}})
0
```

`0` means success in this case.

### DIFF
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -648,6 +648,8 @@ module.exports = {
     user.storePasswordInHistory();
     user.password = User.hashPassword(newPassword);
     user.lastModified = new Date();
+    user.lastPasswordReset = new Date();
+
     await user.save().catch((err) => {
       logger.error(
         `[UserController->updatePassword] ${err.message}`,


orakili (approved):
Looks good and the commands worked as expected.

PR #313, opened 2021-10-26
HID-2300: allow confirmed recovery email addresses to be used for logins

rupl:
# HID-2300

Allowing confirmed recovery addresses to be used for login will solve a ton of support problems, in addition to allowing simplification of language across the website. There's no additional security risk because the user still has to demonstrate ownership of the email address before it can be used for login.

Per usual, there's no specific feedback in regards to why your login gets rejected. It all gets condensed into one canned response, in order to comply with OICT and follow the recommendations of our recurring independent security audits.

## Testing

```sh
npm run e2e -- -t Login
```

There are new tests which attempt the following scenarios:

- non-existent address: PREVENT login
- unconfirmed recovery used to login: PREVENT login
- confirmed recovery used to login: ALLOW log in

To get these working, your test user should have a profile that looks like the following screenshot. I tried to name the emails as plainly as possible to make it obvious! All addresses end with `example.com`:

Screen Shot 2021-10-26 at 12 22 41

### DIFF
--- a/_tests/e2e/Login.test.js
+++ b/_tests/e2e/Login.test.js
@@ -37,7 +37,6 @@ describe('Login', () => {
 
   it('allows user to log in when credentials are valid', async () => {
     await utils.login(page);
-    await page.waitForSelector('.t-page--dashboard');
 
     expect(await page.url()).toContain('/user');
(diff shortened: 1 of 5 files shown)


orakili (approved):
Working well.

PR #318, opened 2021-10-29
HID-2305: email template updates for Content Squad

rupl:
# HID-2305

Changing copy of many of the templates plus dropping two more unused ones.

## Testing

- Register new account
  - 2 templates: `register`, `post_register` — EN only; user cannot select `locale` during registration
- Reset password on existing account
  - 2 templates: `reset_password` — EN/FR
- Add email to existing account
  - 4 templates: `email_validation`, `email_alert` — EN/FR

Proof each of the email(s) that each action produces. To find a french-speaking account, query Mongo:

```sh
# shell into DB container
$ docker-exec db mongo local

# now you're on the Mongo prompt. Ask it for the email of a french user:
> db.user.findOne({locale: 'fr'},{email: true})

### CODE REVIEW

File: emails/email_alert/fr/html.ejs [hunk 1]
```diff
@@ -1,11 +1,7 @@
...
-<p>Vous recevez ce message parcequ'une nouvelle adresse email, <%= emailAdded %>, a été ajoutée à votre profil Humanitarian ID.</p>
-
-<p>Si vous avez effectué cette action ou demandé à un manager Humanitarian ID de l'effectuer pour vous, vous pouvez ignorer ce message.</p>
-
-<p>En revanche, si l'adresse email citée ci-dessus ne vous appartient pas, contactez nous immédiatement à info@humanitarian.id !</p>
-
-<p>Cordialement,</p>
-
-<% include ../../includes/footer_fr %>
+<% include ../includes/footer_fr %>
```
orakili, line 7:
```suggestion
<% include ../../includes/footer_fr %>
```

This results in the alert emails not to be sent. **Note**: this is included in https://github.com/UN-OCHA/hid_api/pull/320

orakili (changes requested):
@rupl I created a follow-up PR with text changes for the French templates, the fix for the footer inclusion etc.: https://github.com/UN-OCHA/hid_api/pull/320 as that was too troublesome to do with the `suggestions` system.

A question: 

I couldn't find anywhere in the UI where to change an account's locale, so it looks like the French emails are only there for legacy accounts, is that correct? 

If yes, is it worthy to maintain those French templates?

emails/email_alert/fr/html.ejs
rupl: > I couldn't find anywhere in the UI where to change an account's locale, so it looks like the French emails are only there for legacy accounts, is that correct?

That's correct. It's a vestigial HID Contact feature, but for now we're honoring the setting without offering a place to change it. All new accounts default to English.

rupl: Nice catch here, btw. It would have been a fresh regression. Now it will never see the light of day 👍

rupl:
@orakili I had to resolve a merge conflict involving one line of the `email_alert` messages. In english, the phrase "for login purposes" was removed from the first line. I feel like I edited the french translation correctly, but please take a second look at `email_alert`. Thanks!

orakili (approved):
🚀

PR #319, opened 2021-10-29
HID-2288: show password form again when any pw requirement wasn't met

rupl:
# HID-2288

The password reset process is hopefully a lot smoother now. When you fail to meet all password requirements (complexity or re-use) the error is more specific and it now shows the password reset form, instead of forcing you to generate a new link and start over from the beginning.

When you're successful, it drops you directly on the login screen.

## Testing

Test the process with 2FA/non-2FA.

1. Reset password with regular user. Note the password you choose.
2. Do another password reset. Use SAME password. It should show the following:

Screen Shot 2021-10-29 at 12 52 48

3. Enter a new, valid password. It should show the success message on the login page now:

Screen Shot 2021-10-29 at 12 53 47

### DIFF
--- a/_tests/e2e/PasswordReset.test.js
+++ b/_tests/e2e/PasswordReset.test.js
@@ -67,22 +67,20 @@ describe('PasswordReset [no-ci]', () => {
     expect(confirmInvalid).toBeGreaterThan(0);
   });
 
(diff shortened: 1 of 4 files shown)


left23 (approved):
@rupl Looks good, testing instructions and on-screen instructions are clear.

PR #323, opened 2021-11-11
HID-2279: implement password dictionary test whenever passwords are being chosen

rupl:
# HID-2279

The PR implements cracklib via node.js directly inside the API container. When choosing a new password we now compare to a dictionary which also does some normalization to convert 0 to o and other similar substitutions that don't provide much security to dictionary words. It has the option to compare to custom strings in addition to its dictionary and we're passing in the following successive strings:

- Given name
- Family name
- Each email address on the profile (regardless of confirmation status)

## Setup

@teodorescuserban tagging you to see if I did the docker stuff correctly. I believe it's good, because I deployed this branch to dev and everything looks like it works, including logs in ELK for dev env. Still, your 👀  are appreciated.

@orakili theoretically, rebuilding the API container will incorporate all dependencies. However if the thing crashes when you test it, I worked around this by manually running each command in the Dockerfile inside the API container. If the node.js process starts without crashing, you're in business.

## Testing

Reset the password for a user and try to create a couple weak passwords:

- Name1234567!
- user@example.com1

Also try some strong ones, e.g. from a password manager's generator.

No change to the information we disclose to the end user (via API or Auth UI). All new info about failures is log-only.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,8 +4,8 @@ WORKDIR /srv/www
 COPY . .
 
 RUN cp run_node /etc/services.d/node/run && \
-    apk add cracklib && \
-    apk add --virtual .build-deps python3 python3-dev build-base cracklib-dev && \
+    apk add cracklib cracklib-dev && \
+    apk add --virtual .build-deps python3 python3-dev build-base && \
     npm install && \
     npm run docs && \
     apk del .build-deps && \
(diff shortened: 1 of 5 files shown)


teodorescuserban (approved):
Looks great docke-wise, @rupl 
🎖️

orakili (approved):
Tested and working well.

PR #328, opened 2021-11-30
HID-2281: remove IASC special-case

rupl:
# HID-2281

The PR itself could have been tiny, but in the process of debugging and testing I realized there's a whole set of operations relating to this process that don't log any activity. So I added some logs too.

## Testing

1. Log into a local Drupal website of your choice using local HID. The login must succeed so that you can see your username inside the Drupal website.

If you get GuzzleHttp errors, you probably need to add an entry to `/etc/hosts` to point `hid.test` to the container's "external network" IP. I normally find that by running: `/sbin/ip route|awk '/default/ { print $3 }'`

2. Log into HID as an admin, and edit the `iasc-dev` client to note all the credentials, plus add a redirect URL of your choice that points to the local Drupal setup from (1).
3. With the credentials you found in (2), reconfigure any local Drupal to use the IASC credentials.
4. Log into local Drupal with local HID **with the same HID user as you used during step (1)**.

If you are able to log in with the same user in (1) and (4) it seems like a good indication that dropping this code won't disrupt anything.

The additional logging will hopefully illuminate the process as you log in the two times.

### DIFF
--- a/api/services/JwtService.js
+++ b/api/services/JwtService.js
@@ -1,17 +1,41 @@
-const jwt = require('jsonwebtoken');
-const fs = require('fs');
-const rsa2jwk = require('rsa-pem-to-jwk');
-
 /**
  * @module JwtService
  * @description JSON Web Token Service
  */
+
+const jwt = require('jsonwebtoken');
+const fs = require('fs');
+const rsa2jwk = require('rsa-pem-to-jwk');
+const config = require('../../config/env');
+
+const { logger } = config;
+
 module.exports = {
 
   // Generates a token from supplied payload
   issue(payload) {
     const cert = fs.readFileSync('keys/hid.rsa');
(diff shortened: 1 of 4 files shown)


api/services/JwtService.js
```diff
@@ -32,10 +56,9 @@ module.exports = {
 
   generateIdToken(client, user, scope, nonce) {
     const now = Math.floor(Date.now() / 1000);
-    let sub = user._id;
-    if (client.id === 'iasc-prod' || client.id === 'iasc-dev') {
-      sub = user.email;
-    }
```
rupl, line 38: Removing these lines was the specific fix described in the ticket. The rest is logging.

orakili (approved):
After battling with an unrelated error: https://humanitarian.atlassian.net/browse/HID-2320, I could finish the tests and that works as expected.

PR #330, opened 2021-12-01
HID-2290: explain the User.expires property in the model

rupl:
# HID-2290

Docs only! we had a ticket to delete this stuff and only after digging in I realized it's still used. New docs to explain it to future someone.


orakili (approved):

orakili:
If only all the PRs could be like that!

PR #332, opened 2021-12-06
HID-2318: drop all legacy Notification-related API surface

rupl:
# HID-2318

⚠️  branch is based off of #331 and will rebase to `dev` once it is merged ⚠️ 

Another chunk of unused code that somehow passed me by until I was doing cleanup in #331. This is mainly a directory of templates whose controller is long gone, plus a couple attributes and a DB-cleanup script.

## Testing

```sh
# start the mongo CLI

# before migration (the exact value is unimportant)
> db.user.count({'createdBy': {$ne: null}})
15816
> exit

# run the migration
$ docker-compose exec api node commands/migrateCreatedBy.js
... it will produce a success log entry ...

# go back to the mongo CLI

# after migration it should say zero
> db.user.count({'createdBy': {$ne: null}})
0
```


orakili (approved):
Looks good.

PR #333, opened 2021-12-10
HID-2259: clean up UserController.create

rupl:
# HID-2259

A general refactor to meet the following goals:

- Make the code linear and understandable.
- Validate the input more strictly, and provide more consistent feedback when returning HTTP 400s.
- Remove legacy attributes and legacy workarounds for long-past problems.

## Testing

- `doco exec api npm i && doco restart api`
- Attempt to create users via API call and look for 200/400 responses. If you can produce a 500, this PR needs work.
- Try anything you want: malformed emails, missing props, additional props in the object, go nuts!
- Use the website to perform a password reset and ensure the password dictionary still works. I had to refactor the function from a mongoose "method" to a "static" so that it could be run on a not-yet-created user object.


api/services/EmailService.js
```diff
@@ -75,13 +75,14 @@ function send(options, tpl, context) {
...
-  sendRegister(user, appVerifyUrl) {
+  sendRegister(user) {
     const mailOptions = {
       to: user.email,
       locale: user.locale || 'en',
     };
     const hash = user.generateHashEmail(user.email);
-    let resetUrl = addUrlArgument(appVerifyUrl, 'id', user._id.toString());
+    const baseUrl = `${process.env.APP_URL}/verify`;
+    let resetUrl = addUrlArgument(baseUrl, 'id', user._id.toString());
```
rupl, line 85: @orakili see here regarding concerns in the ticket ☝️

api/controllers/UserController.js
```diff
@@ -35,11 +35,6 @@ module.exports = {
    *           given_name:
    *             type: string
    *             required: true
-   *           app_verify_url:
-   *             type: string
-   *             required: true
-   *             description: >-
-   *               Should correspond to the endpoint you are interacting with.
```
rupl, line 42: @orakili second ping regarding ticket comments

orakili (approved):
Tested and only saw the expected 200/400 and apparently no issue with the dictionary. 👍

PR #347, opened 2022-04-05
[HID-2320] prevent errors during OAuth dialog confirmations

rupl:
# HID-2320

Two investigations indicate that the 'lazy' mode of `yar` was the source of problems within our `oauth2orize` plugin, leading visitors who confirm an OAuth connection to see an error some of the time.

[We forked the library](https://github.com/UN-OCHA/yar) and are using it as our dependency while we await a resolution to the upstream PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2509,9 +2509,8 @@
       }
     },
     "@hapi/yar": {
-      "version": "10.1.1",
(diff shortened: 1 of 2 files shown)


rupl:
@orakili this is mostly an FYI to you. We've standardized perms on the fork and left some additional notes to indicate that it's a fork used by infra.

PR #354, opened 2022-04-13
[OPS-7830] Random guess at syntax that might make the Redis TLS option conditional.

cafuego:
Hopefully the test suite will whine if this is not valid?

See also https://github.com/UN-OCHA/tools-ansible/pull/836

### DIFF
--- a/config/web.js
+++ b/config/web.js
@@ -72,6 +72,7 @@ module.exports = {
             host: process.env.REDIS_HOST || 'redis',
             port: process.env.REDIS_PORT || 6379,
             db: process.env.REDIS_DB || '0',
+            tls: process.env.REDIS_TLS === 'true' ? {} : null,
           },
         },
       },


rupl:
I dug into this and while I'm ok just trying it on dev, the docs I found make me feel like it can't be an empty object.

The [catbox-redis API docs](https://github.com/hapijs/catbox-redis/blob/master/API.md#options) (toward the bottom of the section) say the configuration for `tls` should use `ioredis` options.

[ioredis just lists a `ConnectionOptions` type](https://github.com/luin/ioredis/blob/df04dd8/lib/connectors/SentinelConnector/index.ts#L43). I dug through their utils and also found a [TLS interface](https://github.com/luin/ioredis/blob/df04dd8d87a44d3b64b385c86581915248554508/lib/utils/index.ts#L244-L248) whose properties seem similar to the stuff we already define to make the unencrypted connection. So that doesn't make much sense (unless they SHOULD be different? That might be something @cafuego has insight on.)

Finally, some public-facing [ioredis docs](https://docs.redis.com/latest/rs/references/client_references/client_ioredis/#tls) claim a totally different set of config properties, including private/public keys. ugh.

I think for now I'll rename the branch to the `HID-*` pattern, and try deploying a couple times to dev so we can avoid merging bad attempts into our trunk.

teodorescuserban (approved):
hopefully it works. looks good otherwise.

rupl:
I made a branch [OPS-7830](https://github.com/UN-OCHA/hid-api/tree/OPS-7830) that automatically builds and pushes to ECR so we can deploy to dev and test all our haxx.

cafuego:
Cool. yeah, the ioredis stuff seems to think you need to specify all the certs and keys, the stackexchange stuff seems to think not. PhPRedis seems happy enough without keys, so your guess is as good as mine.

cafuego:
I'm setting this to blocked, as the shared redises are going away. They make the Drupals *slower* and we won't need instances that size just for HID sessions.

rupl:
@cafuego should we close this? Last comment is half year ago and seemed to lean in that direction.

cafuego:
Drupal no longer uses redis, so I don't care how they connect to it ;-)

PR #356, opened 2022-04-18
[OPS-7830] Turns out the docker env is passed as a string boolean with a capital T, so check for that.

cafuego:
Turns out `{}` works fine as a TLS config object with the AWS certificate and chain, as evidenced by the `dev` site right now. It may not work so well with self-signed certs. 

Also, it does not work at all with the conditional I added. When set to `null`, the tls config will cause the API to just outright throw 502 Nope. So it will need work, possibly a conditional cache config export based on the redis TLS env variable. Yay, wheee!

### DIFF
--- a/config/web.js
+++ b/config/web.js
@@ -72,7 +72,7 @@ module.exports = {
             host: process.env.REDIS_HOST || 'redis',
             port: process.env.REDIS_PORT || 6379,
             db: process.env.REDIS_DB || '0',
-            tls: process.env.REDIS_TLS === 'true' ? {} : null,
+            tls: process.env.REDIS_TLS === 'True' ? {} : null,
           },
         },
       },


lazysoundsystem (commented):
I was waiting to get hid-api set up properly before reviewing this...

Checking out this branch does give me a 502 - I'm not sure from the ticket whether this is a good thing or not. Checked out the cafuego/ops-7830-hid-redis-tls branch too and that also gives me a 502.

Is this helpful information?

cafuego:
Unless redis is configured with TLS, the config object must not have (even an empty) `tls:` key.

cafuego:
I'm setting this to blocked, as the shared redises are going away. They make the Drupals *slower* and we won't need instances that size just for HID sessions.

rupl:
@cafuego same as other PR, should we close this?

cafuego:
Uh, sure?

PR #359, opened 2022-05-03
chore(release): 5.1.4

lazysoundsystem:


rupl (changes requested):
This seems to be missing the "legacy" requirement that `main` wasn't merged in beforehand. Our docs no longer ask this of you, but since this is the first release since we updated the docs, we need it to happen one last time.

Because it needs to be merged **before** we run `npm run release` I think a new branch will be the simplest fix.

PR #374, opened 2022-08-22
build: Submodules are a bad idea and make many people very angry.

cafuego:
So we turn them off, which will fail the build if someone tries to sneak one in. Win-win!


lazysoundsystem (approved):
I'm not one of the people they make angry - but I do remember trying to use them and ending up disappointed.

PR #377, opened 2022-09-22
[HID-2361] correct paths for alert include within settings pages

rupl:
# HID-2361

Fixes the include paths for a few settings pages. The "delete account" page is for admins only, so very low impact on prod. The other page does indeed prevent 2FA from being enabled on all environments.

@lazysoundsystem you can go ahead and give this a YOLO and we'll test it on dev, where the problem is currently reproducible.

### DIFF
--- a/templates/settings-delete.html
+++ b/templates/settings-delete.html
@@ -28,7 +28,7 @@
 
           <% if (user.is_admin) { %>
               <%-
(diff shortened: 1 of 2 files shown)


lazysoundsystem (commented):
Approval!

PR #387, opened 2022-11-23
[HID-2372] new endpoint to force-reset user passwords

rupl:
# HID-2372

Endpoint force-resets a user's password to something random. The admin never has access to the account, and it can't ever see the password either.

Afterwards, the target user must perform a successful password reset via email link in order to regain access to the account.

## Testing

1. Generate new docs by running `npm run docs` inside container.
2. Generate a JWT for an admin within your local DB.
3. `POST /api/v3/user/{id}/password-admin` via [local swagger](http://hid.test:8080/docs/#/user/post_user__id__password_admin)

Additional confirmation can be observed by logging into HID successfully beforehand with the target user, then confirming that the same password no longer works after the reset (UI won't say why, but logs will specifically confirm wrong password).

### DIFF
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -822,7 +822,7 @@ module.exports = {
     });
 
     logger.info(
-      `[UserController->updatePassword] Admin forced a password reset`,
+      '[UserController->updatePassword] Admin forced a password reset',
       {
         request,
         security: true,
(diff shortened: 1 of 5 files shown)


lazysoundsystem (approved):
Still having trouble getting this to build locally:
`make: *** No rule to make target 'Release/obj.target/cracklib/src/cracklib.o', needed by 'Release/obj.target/cracklib.node'. Stop.`
but I looked through the code and it seems okay.

orakili (approved):
Sorry for the late review. I could test locally and it works as described.

PR #394, opened 2023-02-09
chore(ci): migrate from Travis to GHA

rupl:
Refs: OPS-9006


rupl:
Wow that's lightning fast! ⚡

PR #399, opened 2023-03-29
chore: attempting to upgrade node.js

rupl:
Refs: HID-2382

Apparently the case matters for auto-building branches.


teodorescuserban:
Looks good. Shame on whoever set only capital HID in the config.

rupl:
@teodorescuserban a remaining known issue is that I couldn't update `package-lock.json` because the `dev` image is still on node-14 and I had trouble rebuilding locally (details noted in the ticket).

Should I just merge, let the thing build, pull locally, and try again?

rupl:
heh, I deployed this to dev using the auto-build tag, but I need to log the node version to ELK for it to be of any use.

PR #401, opened 2023-04-03
[HID-2382] node.js upgrade part 2

rupl:
# HID-2382

Previous PR for this ticket #399

In that branch I'd had some local troubles which I since resolved. With a totally restored local environment, I was able to address the peerdeps failures that npm 9 is stricter about. We had to upgrade many `eslint` and `jest` related packages, but both sets of tests ran successfully on my local. The PR will do linting automatically as a second opinion. If it's green, I'll merge.


rupl:
hmmm. I guess I need to adjust some rules after all

PR #405, opened 2023-05-12
[HID-2387] user counts

rupl:
# HID-2387

Admin-only endpoint for aggregate user counts.

### Testing

- `GET /api/v3/numbers`
- No token or non-admin token should 403
- Bad token should 401
- Admin token should 200 with a JSON object:

```json
{
  "totalUsers": 321,
  "emailVerified": 123,
}
```

### DIFF
--- a/api/controllers/NumbersController.js
+++ b/api/controllers/NumbersController.js
@@ -1,0 +1,65 @@
+/**
+ * @module NumbersController
+ * @description Provides aggregate counts for: users.
+ */
+const Boom = require('@hapi/boom');
+const User = require('../models/User');
+const config = require('../../config/env');
+
+const { logger } = config;
+
+module.exports = {
+
+  /*
+   * @api [get] /numbers
+   * tags:
+   *   - numbers
+   * summary: Returns total number of users in the database.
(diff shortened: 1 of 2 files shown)


api/controllers/NumbersController.js
```diff
@@ -0,0 +1,65 @@
+/**
+ * @module NumbersController
+ * @description Provides aggregate counts for: users.
+ */
+const Boom = require('@hapi/boom');
```
cafuego, line 5: 💥

cafuego (approved):
Assuming I can make auth for for n8n, this ought to do the trick for reporting.

orakili (approved):

orakili:
Note:  approved based on the code review but I'm still having unrelated issues that prevent me from testing locally.

rupl:
That's ok, we can put it on dev and test it. I tried to set up the unit tests for the basics: missing, invalid, or non-admin tokens. We saw how that PR is going, but it might be fixable without adding a DB export to CI.

@cafuego I think for n8n the easiest thing to do it to make a bot account. More details in ticket.

PR #406, opened 2023-05-12
[HID-2028] set up unit tests

rupl:
# HID-2028

Groundwork for unit testing the Controllers. Without a DB it can only test parts of the app, but I thought I would throw it onto GitHub and see how it works inside the GHA.

Basically all of the "failure" test-cases from #405 are contained in this branch. I didn't want to go overboard writing tests; just enough to see if it works and to test our new code.


orakili (approved):
Code looks good to me. The github action to run the tests failed. Any opinion on that?

rupl:
My opinion is that it needs work :) I asked for review before seeing if CI would succeed.

rupl:
Ok I promise that wasn't just scope creep. The commands all referenced the old `index.js` assigning it to a variable `app` in most cases. Moreover the statements that used that file were weirdly coded anyway. I've tried to standardize on something that makes a bit more sense.

PR #407, opened 2023-05-19
[HID-2388] CD v8

rupl:
# HID-2388

- Upgrades HID to CD v8.1.0
- Replaced the old `cd.css` Sass artifact with `assets/css/hid.css` and a mini-build using PostCSS that imports and then minifies, nothing else. Run with `npm run css` and the results have to be checked into version control to avoid build steps on various environments.
- Found a better way to port the Drupal Behaviors files to non-Drupal, so the files can be more easily updated in the future.
- `cd-overrides.css` contains all the site-specific customizations, just like WP themes.

**No need for local setup!** You can view the work on dev (compare to prod): https://dev.auth-humanitarian-id.ahconu.org/ — if it doesn't look different then we can re-run Jenkins:

- Job: `hid-dev-deploy`
- STACK: `dev`
- IMAGE_TAG: `HID-2388`


left23 (approved):
Looks good!

PR #408, opened 2023-05-24
[HID-2389] DRY session auth

rupl:
# HID-2389

Replaces the boilerplate conditionals in each route handler with an authentication strategy specified on each route. The ticket has more details as to why I wrote a custom plugin. TL;DR we have technical/business reasons to hold onto `yar` so the custom plugin was the smallest change that leverages the existing codebase.

End result in the controllers is that the `cookie` is guaranteed to exist by the time the route handler is executing. We no longer need all the conditionals and redirects that used to be at the top of every route handler.

### DIFF
--- a/config/web.js
+++ b/config/web.js
@@ -13,6 +13,7 @@ const hapiRateLimit = require('hapi-rate-limit');
 const oauth2orizeExt = require('oauth2orize-openid');
 const hapiOauth2Orize = require('../plugins/hapi-oauth2orize');
 const hapiAuthApi = require('../plugins/hapi-auth-api');
+const hapiAuthSession = require('../plugins/hapi-auth-session');
 const Client = require('../api/models/Client');
 const OauthToken = require('../api/models/OauthToken');
 const JwtService = require('../api/services/JwtService');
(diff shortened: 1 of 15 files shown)


config/web.js
```diff
@@ -111,7 +112,7 @@ const config = {
 
         // Configure how cookies behave in browsers.
         cookieOptions: {
-          password: process.env.COOKIE_PASSWORD,
+          password: process.env.COOKIE_PASSWORD || 'testtesttesttesttesttesttesttest',
```
orakili, line 115: Can you confirm the default is expected here and not a leftover?

orakili (approved):
Code looks ok. Couldn't test locally.

config/web.js
rupl: I did do that on purpose. But now that you ask, I suppose in this particular case we'd prefer to default to something random, juuuust in case it makes it to production. During CI runs, the value doesn't matter because cookie doesn't have to last longer than a unit test (curiously, exporting the `COOKIE_PASSWORD` in the Action didn't seem to work if you look at a couple of the [failed runs](https://github.com/UN-OCHA/hid-api/actions/runs/5080550178/jobs/9127663778)).

PR #409, opened 2023-05-25
[HID-2391] Branded 404 page

rupl:
# HID-2391

I set up a catch-all route that looks at the `Accept` header to see if `text/html` is accepted, and returns a branded 404 in that case. Otherwise it will return the existing JSON response `Boom.notFound()`


orakili (approved):
Approving based on the code review. Still cannot test locally.

PR #433, opened 2024-04-11
[HID-2413] CD v9.4.0

rupl:
# HID-2413

Upgrades to [CD v9.4.0](https://github.com/UN-OCHA/common_design/releases/tag/v9.4.0). Notable adjustments:

- OCHA Services markup is updated, and no longer gets moved into the Global Header via JS. It appears in the footer.
- CD Dropdown is updated with a11y and UX improvements.
- CD Alert is updated with many a11y improvements.
- Landmark roles now have labels instead of duplicate `role` attributes.

Additional adjustments that were HID-specific:

- Did NOT apply the `preload` markup for fonts as the inline JS required the CSP to be adjusted in such a way as to render it toothless for JS.
- Removed IE11 polyfills


orakili (commented):
Sorry for the late review. I tested the upgrade and everything looks fine. Thanks.

rupl:
Thanks!

PR #434, opened 2024-04-12
feat: Naive attempt to add SES role based auth for emails, instead of SMTP.

cafuego:
Perhaps this should be in config, and pass a single mail transport back to the module.

Refs: OPS-10025

### DIFF
--- a/api/services/EmailService.js
+++ b/api/services/EmailService.js
@@ -5,20 +5,42 @@
 const Nodemailer = require('nodemailer');
 const Email = require('email-templates');
 const config = require('../../config/env');
+const aws = require("@aws-sdk/client-ses");
+const { defaultProvider } = require("@aws-sdk/credential-provider-node");
 
 const { logger } = config;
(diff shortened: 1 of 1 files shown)


rupl:
I'd tend to agree it should be in config, but I also don't have so strong of an opinion. I added the deps and linted for you. Should we deploy this to dev and see if it works? Is it even possible to test when mailhog is scooping up emails?

cafuego:
Sadly, no. There is no way to test this change with mailhog. When mail is handled via SES api calls, the data is handed off to an AWS service endpoint and the only thing we know is the API response code and the (eventual) log in ELK. There is no way to intercept that.

To make things worse, the there is no test endpoint to do a dry run with; to check whether or no the API calls work or fail there are magic destination email address that will trigger a success or fail. Any other email addresses will cause an actual email to be delivered.

That means we cannot test this with an actual HID database; at best we can spin up an env with only our own accounts  in it and make it send mail.

